PR #2, opened 2018-08-09
reorganize navigation to use categories

mojavelinux:
Shift all the navigation up one level to take advantage of the styling of categories in the new docs UI. The following screenshots shows the difference in appearance.

Before:

before

After:

after

The new docs UI expands the top-level categories by default, at least for now.

### DIFF
--- a/modules/nav.adoc
+++ b/modules/nav.adoc
@@ -1,31 +1,39 @@
+.Introduction
 * xref:ROOT:index.adoc[What's New]
 * xref:ROOT:getting-started.adoc[Getting Started]
-* Access Control
-** xref:ROOT:authentication.adoc[Authentication]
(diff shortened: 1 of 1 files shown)


mojavelinux:
Note that if you were to move the navigation file inside the ROOT folder, you could drop the `ROOT:` segment from each entry. To avoid conflicts, that change will need to be made in a follow-up PR.

jamiltz:
Great to know about that. Ok I'll remove `ROOT:` after merging this PR. Thanks!

jamiltz:
Done in https://github.com/couchbaselabs/docs-sync-gateway/commit/d16910814229fadfaf1799029d2f7e223f7f6462. Output looks good.

PR #3, opened 2018-08-09
escape unintentional formatting and attribute references

mojavelinux:
- enclose codespans that contain syntax that matches an attribute reference in a passthrough
- escape unconstrained formatting marks in URLs


jamiltz:
Thanks!

PR #4, opened 2018-08-17
clean AsciiDoc (master)

graphitefriction:
- fix titles and remove extra lines
- remove quote blocks and add admonitions
- remove indents/spaces from tables and add implicit headers
- remove unconstrained inline formatting
- simplify lists
- remove unneeded escapes
- remove attribute files
- convert link:$$$.html to xref:$$$.adoc


jamiltz:
@graphitefriction those are wonderful changes!

One question, I noticed this PR adds explicit title ids. For example:

```
[#custom-auth]
== Custom Authentication
```

I'm wondering if it's really needed here?

jamiltz:
My preference would be to auto-generate the title IDs with hyphens instead of underscores. So I use the following on my local Antora config to do that:

```
asciidoc:
  attributes:
    idprefix: ''
    idseparator: '-'
```

graphitefriction:
@jamiltz I can definitely remove the explicit IDs (I added them out of habit because I tweak section titles constantly in my own writing and hate grepping through all the rest of the pages to fix the links I broke :stuck_out_tongue_closed_eyes: ) I agree 100% with putting the AsciiDoc attributes in the playbook. @mojavelinux is concerned that there may be a few ID collisions when applied across the whole migration set, so he's doing a test run. If there aren't massive conflicts, we'll add the attributes to the production and staging playbooks and remove the attributes from the document headers (DRY FTW!)

jamiltz:
Sounds good. Thanks! :)

graphitefriction:
@jamiltz I removed the explicit IDs and updated the PR. For now, the id attributes are in the headers of the targeted documents. Dan's got a big N1QL reference document with colliding IDs he has to debug before the attrs can be added to the playbook.

jamiltz:
Ok!

mojavelinux:
Thanks James!

PR #10, opened 2018-08-23
rewrite tabsets to use new structure supported by extension (master)

mojavelinux:
A tabset is defined using a description list (dlist) wrapped in an example block. Each term becomes the tab's title and the blocks attached to the term (aka the details) become the tab's content. (The open block surrounding the attached blocks is optional).

```
[tabs]
====
Tab A::
+
Content for Tab A.

Tab B::
+
--
Content for Tab B.

More content for Tab B.
--
====
```

We annotated the block as follows to avoid warnings when the block extension is not registered (such as in a preview tool).

```
:tabs:

[{tabs}]
====
....
====
```


jamiltz:
Ok, updated installation section on couchbase lite docs to use the same syntax.

mojavelinux:
:+1:

mojavelinux:
I'll make this live later today.

jamiltz:
Great!

PR #13, opened 2018-08-31
Reverse "use_views" bool values in compatibility matrix

bbrks:
`use_views: false` is only compatible with >=5.5
`use_views: true` is compatible with all current versions

I've reversed the values in the compatibility matrix to correct this.


jamiltz:
Indeed. Good catch!

PR #17, opened 2018-10-11
Doc 4260

jamiltz:


modules/ROOT/pages/shared-bucket-access.adoc
```diff
@@ -95,6 +95,20 @@ Document operations also have a different impact on tombstones when Shared Bucke
 |Removes the tombstone metadata
 |===
 
+== FAQ
+
+=== How do I query for all documents in a channel with N1QL?
+
+// TODO
```
jamiltz: @adamcfraser do we have an example N1QL statement to answer this question?

rajagp: Please check : https://forums.couchbase.com/t/n1ql-sync-gateway/18427/2

I would also add the statement WARNING user that the sync metadata structure could change at any point and applications should not rely on that to drive their business logic. This is a convenience for introspecting their document sync metadata.

jamiltz: Ok thanks!

modules/ROOT/pages/shared-bucket-access.adoc
```diff
@@ -95,6 +95,20 @@ Document operations also have a different impact on tombstones when Shared Bucke
...
+
+=== How do I access a blob stored in Couchbase Lite?
+
+See the xref:couchbase-lite:ROOT:swift.adoc#blobs[Blob] example.
+
+=== How do I access an attachment from a WebApp?
+
+Attachments can be accessed through Sync Gateway's REST API using the xref:rest-api.adoc#/attachment/get\__db___doc___attachment_[/{db}/{doc}/{attchament}] endpoint.
```
jamiltz: @rajagp are those 2 questions above sufficient or should we cover more? I don't know if read/write of blobs is supported through the SDK.

rajagp: Pleas check w. @adamcfraser  on other gotchas to be covered.
No- we don't support blogs through SDK. Convergence is for docs only at this point - It will be handled in a future release

PR #21, opened 2018-11-20
DOC-4377

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4377

### DIFF
--- a/modules/ROOT/assets/attachments/sync-gateway-admin.yaml
+++ b/modules/ROOT/assets/attachments/sync-gateway-admin.yaml
@@ -189,7 +189,7 @@ parameters:
     name: open_revs
     in: query
     description: |
-      Option to fetch specified revisions of the document. The value can be `all` to fetch all leaf revisions or an array of revision numbers (i.e. open_revs=["rev1", "rev2"]). Only [leaf revision](glossary.html) bodies are guaranteed to be returned.
+      Option to fetch specified revisions of the document. The value can be `all` to fetch all leaf revisions or an array of revision numbers (i.e. open_revs=["rev1", "rev2"]). Only [leaf revision](glossary.html) bodies that haven't been pruned are guaranteed to be returned.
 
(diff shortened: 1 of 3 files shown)


modules/ROOT/assets/attachments/sync-gateway-admin.yaml
```diff
@@ -188,7 +188,10 @@ parameters:
   open_revs:
     name: open_revs
     in: query
-    description: 'Option to fetch specified revisions of the document. The value can be `all` to fetch all leaf revisions or an array of revision numbers (i.e. open_revs=["rev1", "rev2"]). If this option is specified the response will be in multipart format. Use the `Accept: application/json` request header to get the result as a JSON object.'
+    description: |
+      Option to fetch specified revisions of the document. The value can be `all` to fetch all leaf revisions or an array of revision numbers (i.e. open_revs=["rev1", "rev2"]). Only [leaf revision](glossary.html) bodies are guaranteed to be returned.
```
tleyden: Leaf revisions aren't quite guaranteed to be returned because of the rev tree pruning algorithm, which is controlled by the `maxRevTreeDepth ` setting.  The [pruning](https://blog.couchbase.com/database-sizes-and-conflict-resolution/) section of this blog post describes that.

adamcfraser: @tleyden Are you talking about tombstone leaf revisions that have been pruned?  Strictly speaking if they've been pruned they are no longer leaf revisions.

tleyden: Yeah it might be worth clarifying that point.  "Only leaf revision bodies that haven't been pruned are guaranteed to be returned".

Also, with XATTRS, isn't there another caveat around tombstoned leaf revisions that were deleted outside the metadata purge interval?

jamesnocentini: Updated per review. @adamcfraser can you confirm Traun's point?

PR #22, opened 2018-11-20
DOC-4435

jamesnocentini:

### DIFF
--- a/modules/ROOT/pages/glossary.adoc
+++ b/modules/ROOT/pages/glossary.adoc
@@ -7,3 +7,22 @@ Optional feature in xref:couchbase-lite::swift.adoc#handling-conflicts[Couchbase
 No conflicts mode::
 The process by which write operations that would result in a conflict are rejected by the system.
 Optional feature in xref:couchbase-lite::swift.adoc#handling-conflicts[Couchbase Lite 2.0 and above] and xref:config-properties.adoc#databases-foo_db-allow_conflicts[Sync Gateway 2.0 and above].
+
+Mobile Tombstone::
+A tombstone is the record of a deleted item.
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/glossary.adoc
```diff
@@ -6,4 +6,23 @@ Optional feature in xref:couchbase-lite::swift.adoc#handling-conflicts[Couchbase
...
 No conflicts mode::
 The process by which write operations that would result in a conflict are rejected by the system.
-Optional feature in xref:couchbase-lite::swift.adoc#handling-conflicts[Couchbase Lite 2.0 and above] and xref:config-properties.adoc#databases-foo_db-allow_conflicts[Sync Gateway 2.0 and above].
\ No newline at end of file
+Optional feature in xref:couchbase-lite::swift.adoc#handling-conflicts[Couchbase Lite 2.0 and above] and xref:config-properties.adoc#databases-foo_db-allow_conflicts[Sync Gateway 2.0 and above].
+
+Mobile Tombstone::
+A tombstone is the record of a deleted item.
+Tombstones are created to allow all devices to see that a document has been deleted - particularly in the case of devices that may not be online continuously and therefore not syncing regularly.
+The `_deleted: true` property is used internally to identify tombstones.
```
adamcfraser: I think it's worth making clear that a tombstone is still just a revision of a document.  The 'record of a deleted item' terminology sounds a bit like it's something disconnected from the document itself, when in fact a tombstone is just a type of document revision.

I suppose we could consider using the nomenclature 'tombstone revision' everywhere, both to make this clear, and to better distinguish between this and Couchbase Server tombstones?

adamcfraser (changes requested):

modules/ROOT/pages/glossary.adoc
jamesnocentini: Yes good points. I've made those adjustments in the last commit of the PR.

PR #23, opened 2018-11-20
DOC-4435

jamesnocentini:

### DIFF
--- a/modules/ROOT/pages/managing-tombstones.adoc
+++ b/modules/ROOT/pages/managing-tombstones.adoc
@@ -68,7 +68,10 @@ The table below describes those differences.
 2+a|*Tombstone storage location*
 |Document
 |System extended attribute (`xattr._sync`).
-Empty document body.
+
+2+a|*Additional properties on a tombstone*
+|Document
+|N/A
 |===
 
 Document operations also have a different impact on tombstones when Shared Bucket Access is enabled/disabled.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/managing-tombstones.adoc
```diff
@@ -68,7 +68,10 @@ The table below describes those differences.
 2+a|*Tombstone storage location*
 |Document
 |System extended attribute (`xattr._sync`).
-Empty document body.
+
+2+a|*Additional properties on a tombstone*
+|Document
+|N/A
```
adamcfraser: It might be better to list these as 'Supported'/'Not Supported' or 'Persisted'/'Not Persisted' or 'Retained/Not Retained' to make it clear what's going to happen if users try to include data on a tombstone.  I think the additional clarity is necessary because CBL (and the SG REST API) is going to let users wrute a full document body on a delete (along with _deleted:true), but SG isn't going to store that body when shared_bucket_access is enabled.

adamcfraser (changes requested):

modules/ROOT/pages/managing-tombstones.adoc
jamesnocentini: OK. I have included that in https://github.com/couchbase/docs-sync-gateway/pull/23/commits/b7324876257b4bdac3a21961a8c5e5a5f2a3cd23. Decided to spell out the scenario of additional props through CBL/SG REST APIs (when shared bucket access is enabled) but let me know if you'd rather keep that out.

rajagp (approved):

PR #27, opened 2018-11-30
DOC-4496

jamesnocentini:


modules/ROOT/assets/attachments/sync-gateway-admin.yaml
```diff
@@ -655,7 +655,7 @@ paths:
               docs:
                 type: array
```
bbrks: I think this is a bit misleading. The response is a multipart/mixed response of single JSON documents, not an array of them. This is how the current doc renders the response:

screen shot 2018-11-30 at 12 07 33

Here's a real-world response:

```
$ curl -i -X POST http://localhost:4985/db/_bulk_get -H 'Content-Type:application/json' -d '{"docs": [{"id": "doc1234"}, {"id": "doc1235"}]}'
--
HTTP/1.1 200 OK
Content-Type: multipart/mixed; boundary="1cba224ff2aa106566e3ab65de9c861c24558ba368f8cd7f6fcde53b88f4"
Server: /2.5 branch/ commit/ EE
Date: Fri, 30 Nov 2018 11:52:42 GMT
Content-Length: 426
 
--1cba224ff2aa106566e3ab65de9c861c24558ba368f8cd7f6fcde53b88f4
Content-Type: application/json
 
{"_id":"doc1234","_rev":"1-c543d6514c609f65180f94af247aaffe","hello":"world!"}
--1cba224ff2aa106566e3ab65de9c861c24558ba368f8cd7f6fcde53b88f4
Content-Type: application/json; error="true"
 
{"error":"not_found","id":"doc1235","reason":"missing","status":404}
--1cba224ff2aa106566e3ab65de9c861c24558ba368f8cd7f6fcde53b88f4--
```

I mentioned it in the Jira ticket already, but only OpenAPI Spec v3 supports multipart properly. I'm not really sure how we can document the above response, other than maybe plain-text.

bbrks (changes requested):
The response is a bit misleading, as we don't return an array of "docs", but a multipart response of single JSON documents. Take a look at the comment and see if you agree.

modules/ROOT/assets/attachments/sync-gateway-admin.yaml
jamesnocentini: Yes indeed, hadn't considered that option.
Updated the PR. The response in the UI now looks like the following:

image

@bbrks does that look ok?

bbrks:
Cool! Yeah that's much clearer. Thanks!

bbrks (approved):

PR #30, opened 2018-12-05
Doc 4521

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4521

### DIFF
--- a/modules/ROOT/pages/data-modeling.adoc
+++ b/modules/ROOT/pages/data-modeling.adoc
@@ -13,6 +13,14 @@ A document which contains user properties with a leading underscore will be reje
 This rule applies to writes performed through Couchbase Lite SDKs and Sync Gateway REST APIs.
 And to writes performed through Couchbase SDKs when xref:shared-bucket-access.adoc[Mobile-Web Data Sync] is enabled.
 
+This may be an issue in the following scenarios:
+
+* In a deployment with Mobile-Web Data Sync and Field Level Encryption enabled, this rule conflicts with the default xref:java-sdk::encryption.adoc#format[field encryption format].
+* http://ottomanjs.com/[Ottoman.js] adds an `_type` attribute to the root of the document.
+So if you are using Ottoman within your app, then you would encounter this issue.
+
 Note that in a deployment with Mobile-Web Data Sync and Field Level Encryption enabled, this rule conflicts with the default xref:java-sdk::encryption.adoc#format[field encryption format].
+
 === Workaround
+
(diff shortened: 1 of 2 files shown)


jamesnocentini:
@rajagp can you provide a PR review on this one?

rajagp (commented):
I think this looks good. Only thing is that the workaround applies to all scenarios and should be mentioned as such. I think its important to call out all the known issue scenarios (Ottomon and the encryption stuff that you mention) - maybe state "impacted use cases" .

modules/ROOT/pages/data-modeling.adoc
```diff
@@ -0,0 +1,17 @@
...
+The underscore character (`+_+`) is a reserved prefix for system properties of a document (`+_id+`, `+_rev+` etc.)
+A document which contains user properties with a leading underscore will be rejected by Sync Gateway with the following error.
+[source,text]
+----
+"{"error":"Bad Request","reason":"user defined top level properties beginning with '_' are not allowed in document body"}"
+----
+This rule applies to writes performed through Couchbase Lite SDKs and Sync Gateway REST APIs.
+And to writes performed through Couchbase SDKs when xref:shared-bucket-access.adoc[Mobile-Web Data Sync] is enabled.
+
+Note that in a deployment with Mobile-Web Data Sync and Field Level Encryption enabled, this rule conflicts with the default xref:java-sdk::encryption.adoc#format[field encryption format].
```
rajagp, line 22:
```suggestion
This may be an issue in the following scenarios 
- In a deployment with Mobile-Web Data Sync and Field Level Encryption enabled, this rule conflicts with the default xref:java-sdk::encryption.adoc#format[field encryption format].
- Ottoman (ADD LINK) adds _type attributes to the root of the document. So if you are using Ottoman within your app, then you would encounter this issue.

```

modules/ROOT/pages/data-modeling.adoc
```diff
@@ -0,0 +1,17 @@
...
+The recommended workaround in this scenario is to wrap the property prefixed with an underscore in another object whose key isn't prefixed with an underscore.
```
rajagp:
```suggestion
=== Workaround
The recommended workaround in these scenarios is to wrap the property prefixed with an underscore in another object whose key isn't prefixed with an underscore.
```

modules/ROOT/pages/data-modeling.adoc
jamesnocentini: suggestion added in https://github.com/couchbase/docs-sync-gateway/pull/30/commits/1b9f9ca71d6a196e510a5d2b4e2433add76ada80

PR #35, opened 2019-01-15
DOC-4482

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4482


jamesnocentini:
@mattcarabine when able, this doc would probably also need your review for accuracy.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+The following sections cover the steps to deploy a cluster of regular Sync Gateway nodes.
+The configuration and deployment controller files will differ if you are deploying a cluster of import nodes or a single replicator node.
+
+For example, to deploy a cluster with 4 regular nodes, 1 import node and 1 replicator node you would run the steps below 3 times, each time with a different configuration and deployment controller file.
+
+=== Configuration file
+
+In this section, you will create the Sync Gateway configuration file and store it in a secret location.
+
+. Create a new file called *sgw-config.json* with the following and replace the `server` key with the addressable Couchbase Server hostname.
```
mattcarabine: Do we need to call out that in a real scenario, the user would replace this with config appropriate for their deployment?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+      "revs_limit": 20
+    }
+  }
+}
+----
+<1> The server key should point to the Couchbase Server cluster.
+This would typically be of the form `http://CB_SERVER_POD.CB_SERVER_SERVICE_NAME:8091`.
+<2> The username/password keys should match what was setup in the <<pre-requisites, pre-requisites>> section when you configured the Sync Gateway RBAC user.
+. You will use a https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/[Kubernetes configMap] to pass the configuration file to Sync Gateway on launch.
+However, since the Sync Gateway contains sensitive information, it is recommended that you create a secret from the config file and pass that to the Sync Gateway.
```
mattcarabine: It seems a little confusing to discuss configmap AND secret, since while they're similar, you can't create a secret from the configmap.

Perhaps reference this in a way that recommends a secret (and shows the steps below), but then say "you could also use a configmap if you are not concerned about the extra security the secret adds"?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+              mountPath: /sync-gateway-config
+              readOnly: true
+          env:
+            - name: GOMAXPROCS
+              value: "2"
+          resources:
+            requests:
+              cpu: "100m"
+            limits:
+              cpu: "100m"
```
mattcarabine: This is limiting Sync Gateway to 100millicpus (1/10th of a CPU), whereas `GOMAXPROCS` is set for 2 CPUs. Probably want to also update these two to 2 too, as it's unlikely a user would ever want to give such extremely pitiful resources to Sync Gateway.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+A deployment controller allows you to define the number of Sync Gateway replicas and other parameters.
+
+. Create a new file called *sgw-deployment.yaml* with the following.
++
+[source,yaml]
+----
+apiVersion: extensions/v1beta1
+kind: Deployment
+metadata:
+  name: sync-gateway <1>
```
mattcarabine: If you were to copy paste this file as-is, it would be rejected by Kubernetes due to the number included in brackets.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+spec:
+  replicas: 2 <2>
```
mattcarabine: If you were to copy paste this file as-is, it would be rejected by Kubernetes due to the number included in brackets.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+  template:
+    metadata:
+      labels:
+        app: sync-gateway
+    spec:
+      containers:
+        - name: sync-gateway
+          image: couchbase/sync-gateway:2.1.2-enterprise <3>
```
mattcarabine: If you were to copy paste this file as-is, it would be rejected by Kubernetes due to the number included in brackets.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+          args: ["/sync-gateway-config/sgw-config.json"] <4>
```
mattcarabine: If you were to copy paste this file as-is, it would be rejected by Kubernetes due to the number included in brackets.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+          volumeMounts:
+            - name: sgw-config-volume
+              mountPath: /sync-gateway-config
+              readOnly: true
+          env:
+            - name: GOMAXPROCS
```
mattcarabine: Don't know if this needs a comment about this limiting Sync Gateway to using 2 CPU?
If this is not set it will create as many processes as CPU cores on the host machine, so is important to bear in mind in a Kubernetes environment.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
++
+[source,console]
+----
+deployment.extensions "sync-gateway" created
+----
+. You can check the status of the deployment with the following command until all the pods corresponding to the Sync Gateway are in the "Ready" state and the status is "Running".
++
+[source,console]
+----
+kubectl get pods
```
mattcarabine: Up to you, but `kubectl` also offers a `--watch` flag, which will keep sending in new updates as they come in, rather than needing to run the command multiple times.

Other Kubernetes tutorials I've read online for other software tend to recommend that approach, although it doesn't make that much difference :)

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+----
++
+If successful, you will see the following.
```
mattcarabine: Do we need to clarify 'success criteria' here? The user may not see the cb-example etc depending on if they've deployed their Couchbase Server in the same namespace.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+Follow these steps to deploy a load balancer in front of the Sync Gateway cluster.
+
+. Create a new file called *sgw-load-balancer.yaml* with the following.
++
+[source,yaml]
+----
+kind: Service
+apiVersion: v1
+metadata:
+  name: sgw-load-balancer <1>
```
mattcarabine: As above with copy+pastable-ness.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+spec:
+  selector:
+    app: sync-gateway <2>
```
mattcarabine: As above with copy+pastable-ness.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+  ports:
+  - protocol: TCP
+    port: 4984 <3>
```
mattcarabine: As above with copy+pastable-ness.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+----
+service "sgw-load-balancer" created
+----
+. Verify the status of the service creation with the following.
++
+[source,console]
+----
+kubectl get services
+----
+If successful, you will see the following.
```
mattcarabine: Again, do we want to point out what the user is looking for (i.e. `sgw-load-balancer`)?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+sync-gateway-7474f5df4b-c29xw        1/1       Running   2          18m
+sync-gateway-7474f5df4b-p98sq        1/1       Running   0          18m
+----
+
+== Deploying a Load Balancer
+
+In a production deployment, you will likely have one or more Sync Gateway nodes fronted by a xref:load-balancer.adoc[load balancer].
+
+You will deploy the load balancer using the https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/[Kubernetes Load Balancer service].
+The load balancer service provides an externally accessible IP address and routes traffic to the right ports in the cluster.
```
mattcarabine: Load balancers only work on Cloud Environments (e.g. AWS, GCP etc), on-prem clusters do not have a built-in solution. 
Do we want to point out that out here so someone testing on-prem doesn't expect this to work. We can direct these to use a service such as `NodePort` or `Ingress` instead, although we don't need to go into detail about how to set these up.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+curl  http://EXTERNAL-IP:4984
+----
+It should return the following.
++
+[source,console]
+----
+{"couchdb":"Welcome","vendor":{"name":"Couchbase Sync Gateway","version":"2.1"},"version":"Couchbase Sync Gateway/2.1.1(17;fea9947)"}
+----
+
+You have successfully deployed a Sync Gateway cluster on Kubernetes.
```
mattcarabine: One thing I think would be nice here is a quick example of getting the Sync Gateway logs to check it's running correctly (i.e. connected to the bucket, as it might be running but not successfully connecting to Couchbase Server).

This serves 2 purposes:
1. Introduces the user to the concept of retrieving the Sync Gateway logs through Kubernetes-native APIs.
2. It ensures that the user has definitely set up Sync Gateway + CB Server correctly.

mattcarabine (changes requested):
Left some feedback, on the whole looks pretty good, just a few minor suggestions here and there.

Only major issue I had was that none of the file examples are copy-pastable because of the numeric annotations.

From an asciidoc/antora perspective, would these annotations actually be copied or would it just be CSS styled?
If they copy+paste without annotations in then feel free to ignore those comments, they were just what I saw when reviewing via GitHub.

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: How about this ? 
```suggestion
Open the *sgw-config.json* file corresponding to your deployment or create a new file called *sgw-config.json* with the following configuration . In both cases,  replace the `server` key with the addressable Couchbase Server hostname.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+      "allow_conflicts": false,
+      "revs_limit": 20
+    }
+  }
+}
+----
+<1> The server key should point to the Couchbase Server cluster.
+This would typically be of the form `http://CB_SERVER_POD.CB_SERVER_SERVICE_NAME:8091`.
+<2> The username/password keys should match what was setup in the <<pre-requisites, pre-requisites>> section when you configured the Sync Gateway RBAC user.
+. You will use a https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/[Kubernetes configMap] to pass the configuration file to Sync Gateway on launch.
```
rajagp: OK . proposing this 
```suggestion
. You will use a https://kubernetes.io/docs/concepts/configuration/secret/[Kubernetes Secrets] to pass the configuration file to Sync Gateway on launch.
```

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: Agreed. Added a clarification to previous line as well
```suggestion
Alternatively, you could use https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/[Kubernetes configMap] if you are not concerned about the security. However, since the Sync Gateway contains sensitive information, it is recommended that you create a secret from the config file and pass that to the Sync Gateway.
```

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: Good point! Will let @jamiltz  take care of that and come up with a better way of handling inline  references

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz  ^^^

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz  ^^^

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: Can you clarify why  "2" the magic number ?( I think it makes sense to specify what this config does in general . Recommend 2 and then leave it to user.)

*That said*, I think I have to check if we need to restrict SGW pod to a VM or node similar to server- So let me confirm that first. Because if we have anti affinity that we do not need this restriction.

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: Agreed.. I was running out of CPU resources in my limited test env . So suggestion is to keep this same as GOMAXPROCS ?
@jamiltz  please go ahead and switch to 2 and clarify that you align this with GOMAXPROCS

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: OK . Proposing 
```suggestion
kubectl get pods --watch

The --watch option is optional but you use it to be asynchronously notified of  updates to status of the pods instead of having to repeatedly run the command.
```

modules/ROOT/pages/installing-kubernetes.adoc
rajagp:
```suggestion
If successful, you will see a listing of the Sync Gateway pods that were deployed. In the sample output below, we have Couchbase Server and Sync Gateway pods on the same cluster. In a production deployment, you will have Couchbase Server deployed on separate cluster
```

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: Good point. I faced that exact issue with minikube and hence switched to cloud! Didn't know how to get these services going ..probably for another time!
```suggestion
The load balancer service provides an externally accessible IP address and routes traffic to the right ports in the cluster.

NOTE : Load Balancers only work on  Cloud Environments (e.g. AWS, GCP etc). So if you are deploying an on-prem clusters or using something like [minikube](https://github.com/kubernetes/minikube) for your test deployment, this option will not work. Please use a [service](https://kubernetes.io/docs/concepts/services-networking/service/) such  as NodePort or Ingress instead. 
```

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz ^^^ to get rid of the inline reference tags

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz ^^^ to get rid of the inline reference tags

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz ^^^ to get rid of the inline reference tags

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: Yeah, that's what I meant, the actual idea behind the environment variable (rather than the value it's set to, that's pretty arbitrary).

You can add anti-affinity in the deployment, but it is only anti-affinity to the other pods in that deployment. So in theory if you had different deployments with an import node and regular Sync Gateways, these could be scheduled on the same underlying Kubernetes node.
I think it makes sense to try and limit Sync Gateways to running on their own dedicated nodes, but without specifically labelling nodes and ensuring nothing else runs on those besides Sync Gateway, the only way to ensure Sync Gateway 'plays nice' with whatever else is scheduled on that node is to limit its resource and set its GOMAXPROCS (so it knows it has limited resource).

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: 👍

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: 👍

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: Yeah, same as GOMAXPROCS :)

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: 👍

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: Would have to be the same Kubernetes cluster, would just be in a different namespace

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: 👍

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: 👍

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: SGW won't run if it cannot connect to the bucket - as in you will see an error in status. But I can see  value in including a section on Log Collection . For that we should update SGW config to specify the log file path and use kubectl cp to copy files over. 
@jamiltz  For now, just add a subsection under Managing called "Troubleshooting & Log Collection" . I will provide contents of that section separately (using "kubectl cp" and "kubectl logs -f"

modules/ROOT/pages/managing-kubernetes.adoc
```diff
@@ -0,0 +1,24 @@
+= Managing Kubernetes
```
rajagp: I think this should be "Managing Sync Gateway Cluster"

rajagp:
```suggestion
= Managing Sync Gateway Cluster
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+<1> `spec.ports[].targetPort`: the load balancer service targets ports 4984 on the Sync Gateway cluster.
+This is the Sync Gateway port corresponding to the xref:rest-api.adoc[REST API].
+For security purposes, it is recommended that you do not expose the admin port (4985) over the Internet.
+. Deploy the load balancer.
++
+[source,console]
+----
+kubectl create -f sgw-load-balancer.yaml
+----
+If successful, you will see the following.
```
rajagp:
```suggestion
If successful, you will see a listing of the Sync Gateway import pods that were deployed. In the sample output below, we have Couchbase Server and Sync Gateway pods on the same cluster. In a production deployment, you will likely have Couchbase Server deployed on a separate cluster.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,264 @@
...
+In this section, you will create the Sync Gateway configuration file and store it in a secret location.
+
+. Create a new file called *sgw-config.json* with the following and replace the `server` key with the addressable Couchbase Server hostname.
++
+[source,json]
+----
+{
+  "logging": {
+    "console": {
+      "log_keys": ["*"]
```
rajagp:
```suggestion
      "log_file_path": "/var/tmp/sglogs",
    "console": {
      "log_level": "debug",
      "log_keys": ["*"]
    },
    "error": {
      "enabled": true,
      "rotation": {
        "max_size": 20,
        "max_age": 180
      }
    },
    "warn": {
      "enabled": true,
      "rotation": {
        "max_size": 20,
        "max_age": 90
      }
    },
    "info": {
      "enabled": false
    },
    "debug": {
      "enabled": false
    }
      
```

rajagp: @jamiltz  Please include the logging stuff in there in the same log config file

rajagp:
@mattcarabine - Thanks for the in-depth review and feedback. I have addressed all of them  . @jamiltz  Please go ahead and update as suggested . We can do the Troubleshooting and Logging section separately when you are ready for it (Submit a PR for section and I can provide content)

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
              cpu: "2"
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
If successful, you will see a new service corresponding to the load balancer.
In the sample output below, we have the `sgw-load-balancer` service.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: OK. opened a PR for it https://github.com/couchbase/docs-sync-gateway/pull/38

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
If successful, you will see a listing of the Sync Gateway pods that were deployed.
In the sample output below, we have Couchbase Server and Sync Gateway pods running on the same namespace.
In a production deployment, you may have Couchbase Server deployed on a separate namespace.
```

jamesnocentini (commented):
Super thankful for the extensive use of GitHub's suggestions! :) Added a few more.

Re inline references: from my testing, copy-pasting on the docs site excludes the inline references. Example https://docs.couchbase.com/sync-gateway/2.1/shared-bucket-access.html#how-to-enable-it

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: This change got applied somewhere else.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
+= Kubernetes
```
eric-schneider: The page title might be better as "Installing on Kubernetes" to help with SEO, though I'm not sure. I think it helps make the page more identifiable at a glance.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+This page describes the steps to deploy a Sync Gateway cluster on Kubernetes.
+
+== Pre-requisites
```
eric-schneider: OED says it's "prerequisites" with no hyphen, and Google seems to be correcting it too. So I tend to lean toward just a single word.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+
+== Deploying a Sync Gateway Cluster
+
+The Sync Gateway nodes in a cluster are homogeneous in configuration with two exceptions.
+
+* Importer node: under xref:shared-bucket-access.adoc[convergence/shared bucket access], it is recommended that one node handles document import processing.
+For high availability, you can configure more than one node in your cluster to be an import node, though it is strongly discouraged for all Sync Gateway nodes in the cluster to be configured for import processing.
+The configuration of the import Sync Gateway node is slightly different (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
+* Replicator node: if you are using inter-cluster replication using sg-replicate then there will be one designated replicator node whose configuration is different than the rest of the nodes.
+
```
eric-schneider: I think the two list items should stand out more. Maybe bold them or create a definition list?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+* A Couchbase Server RBAC user with application access privileges.
+Refer to xref:getting-started.adoc#creating-an-rbac-user[this section] for instructions.
+
+== Deploying a Sync Gateway Cluster
+
+The Sync Gateway nodes in a cluster are homogeneous in configuration with two exceptions.
+
+* Importer node: under xref:shared-bucket-access.adoc[convergence/shared bucket access], it is recommended that one node handles document import processing.
+For high availability, you can configure more than one node in your cluster to be an import node, though it is strongly discouraged for all Sync Gateway nodes in the cluster to be configured for import processing.
+The configuration of the import Sync Gateway node is slightly different (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
```
eric-schneider: Should this be "The configuration of the Sync Gateway importer node..."?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+* Replicator node: if you are using inter-cluster replication using sg-replicate then there will be one designated replicator node whose configuration is different than the rest of the nodes.
+
+The following sections cover the steps to deploy a cluster of regular Sync Gateway nodes.
+The configuration and deployment controller files will differ if you are deploying a cluster of import nodes or a single replicator node.
```
eric-schneider: Should it be "importer nodes"?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+          resources:
+            requests:
+              cpu: "100m"
+            limits:
+              cpu: "100m"
+      volumes:
+        - name: sgw-config-volume
+          secret:
+            secretName: sgw-config
+----
```
eric-schneider: I noticed that the numbered list doesn't match the number of callouts in the snippet. Maybe I'm missing something

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+spec:
+  selector:
+    app: sync-gateway <2>
+  ports:
+  - protocol: TCP
+    port: 4984 <3>
+    targetPort: 4984
+  type: LoadBalancer
+----
+<1> `metadata.name`: the name of the load balancer is "sgw-load-balancer".
```
eric-schneider: The lists in this guide don't seem to follow a capitalization rule for the first word after a colon. We don't have a style for this yet, though. (I prefer capitalizing the first word after a colon.)

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+<1> `spec.selector.app`: this value corresponds to the pods targeted by the load balancer.
+In this case, it targets any pods with the `app=sync-gateway` label which are the Sync Gateway nodes - this corresponds to what was specified in the deployment yaml file.
+<1> `spec.ports[].targetPort`: the load balancer service targets ports 4984 on the Sync Gateway cluster.
```
eric-schneider: "port" singular

modules/ROOT/pages/managing-kubernetes.adoc
```diff
@@ -0,0 +1,24 @@
...
+----
+kubectl apply -f sgw-deployment-upgrade.yaml
+----
+. You can verify the status of the rollout as follows
++
+[source,console]
+----
+kubectl rollout status deployment sync-gateway
+----
+The deployment controller will ensure that at least one replica is available to server the incoming requests while the other replicas are being upgraded.
```
eric-schneider: "serve"

eric-schneider (changes requested):
I think these are really well-written guides. Well done everyone!

However, I was confused about one bit, and it stems from the fact that you need a different configuration file and deployment controller for each Sync Gateway node type. I couldn't tell where this node type was specified in either the config file or the deployment controller. Perhaps I'm not reading the doc correctly, but if I'm going to go back through these two steps for each node type, I need a little bit more clear information on what the main changes are for each node type. With a normal Couchbase Operator config, you specify the services for each Couchbase node. I was expecting something similar, where I specify the Sync Gateway node type in either the config file or the controller.

Oh, and one other thing related to upgrade. In addition to Sync Gateway version upgrade, customers often upgrade their Kubernetes nodes. You might want to provide some guidance for that scenario. Here is what we have for the Operator: https://docs.couchbase.com/operator/1.1/upgrading-kubernetes.html

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: 👍

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: :+1:

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,297 @@
...
+          volumeMounts:
+            - name: sgw-config-volume
+              mountPath: /sync-gateway-config
+              readOnly: true
+          env:
+            - name: GOMAXPROCS
+              value: "2"
+          resources:
+            requests:
+              cpu: "100m"
```
jamesnocentini:
```suggestion
              cpu: "2"
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
= Installing on Kubernetes
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
== Prerequisites
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
The configuration of the Sync Gateway import node is slightly different (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
The configuration and deployment controller files will differ if you are deploying a cluster of importer nodes or a single replicator node.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
<1> `spec.ports[].targetPort`: the load balancer service targets port 4984 on the Sync Gateway cluster.
```

modules/ROOT/pages/managing-kubernetes.adoc
jamesnocentini:
```suggestion
The deployment controller will ensure that at least one replica is available to serve the incoming requests while the other replicas are being upgraded.
```

jamesnocentini (commented):
Apply feedback as GitHub suggestions.

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/332b254ea676473209694d8cda915e5e1ba8846f

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/3e0c55be4aa347eb81864e6e89eb98f289e02e79

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/6bb09f87009ecc9e14843a0b91bc923f6f278d62

jamesnocentini:
@eric-schneider Thanks for the review! Replies below.

> However, I was confused about one bit, and it stems from the fact that you need a different configuration file and deployment controller for each Sync Gateway node type. I couldn't tell where this node type was specified in either the config file or the deployment controller. Perhaps I'm not reading the doc correctly, but if I'm going to go back through these two steps for each node type, I need a little bit more clear information on what the main changes are for each node type. With a normal Couchbase Operator config, you specify the services for each Couchbase node. I was expecting something similar, where I specify the Sync Gateway node type in either the config file or the controller.

Yes indeed, the next step (after checking with @rajagp) was to include those files.
I've done that in https://github.com/couchbase/docs-sync-gateway/commit/601e2e7b82e5b6d186dbfe230493cc2f408f223e.
The replicator node type is commented out for now because we don't have configs for that yet.

> Oh, and one other thing related to upgrade. In addition to Sync Gateway version upgrade, customers often upgrade their Kubernetes nodes. You might want to provide some guidance for that scenario. Here is what we have for the Operator: https://docs.couchbase.com/operator/1.1/upgrading-kubernetes.html

Done in https://github.com/couchbase/docs-sync-gateway/commit/2381121fce8ea5e9ecc6833d9a6757117e63fe3a

jamesnocentini:
PR ready for merging.
@rajagp do you confirm?

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: Agree. Maybe we could think about a best practices section similar to CBS ( separate PR). 
@jamiltz  Please update text to clarify the purpose of cpus
- cpu : This is used to specify the CPU limit for the Sync Gateway pod. If you do not specify one, the Sync Gateway could spawn as many processes as CPU cores and potentially use up all CPU resources. You can learn more about CPU resource assignment [here](https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#if-you-do-not-specify-a-cpu-limit). We recommend a value of 2 but you should use what is suited for your environment
- GOMAXPROCS : This GO runtime env variable is used to limit the number of system threads that are allocated to the Sync Gateway

modules/ROOT/assets/attachments/kubernetes/sgw-config-import.json
```diff
@@ -0,0 +1,44 @@
+{
+  "logging": {
+    "console": {
+      "log_file_path": "/var/tmp/sglogs",
+      "console": {
```
rajagp, line 5: @jamiltz  Just realized that we should probably set console logging off by default
```suggestion
      "console": {
         "enabled":false
       }
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+This guide demonstrates how to deploy a Sync Gateway cluster on Kubernetes.
+
+== Prerequisites
+
+Before you begin, you must have the following:
+
+* A Couchbase Server cluster already running on Kubernetes.
+If you don't already have one, you can refer to xref:operator::install-kubernetes.adoc[this guide] for instructions.
+* A Couchbase Server RBAC user with application access privileges.
+Refer to xref:getting-started.adoc#creating-an-rbac-user[this section] for instructions.
```
rajagp: @jamiltz  Can you include link that points to the anchor point corresponding to the instructions that show RBAC user creation instead of the top level install guide.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+
+This guide demonstrates how to deploy a Sync Gateway cluster on Kubernetes.
+
+== Prerequisites
+
+Before you begin, you must have the following:
+
+* A Couchbase Server cluster already running on Kubernetes.
+If you don't already have one, you can refer to xref:operator::install-kubernetes.adoc[this guide] for instructions.
+* A Couchbase Server RBAC user with application access privileges.
```
rajagp:
```suggestion
* A Couchbase Server RBAC user with application access privileges. This Sync Gateway will connect to the Couchbase server as this RBAC user.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
+= Installing on Kubernetes
+
+This guide demonstrates how to deploy a Sync Gateway cluster on Kubernetes.
+
+== Prerequisites
+
+Before you begin, you must have the following:
+
+* A Couchbase Server cluster already running on Kubernetes.
```
rajagp: @jamiltz  Not sure why I don't see the link to the guide when I view file....

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+If you don't already have one, you can refer to xref:operator::install-kubernetes.adoc[this guide] for instructions.
+* A Couchbase Server RBAC user with application access privileges.
+Refer to xref:getting-started.adoc#creating-an-rbac-user[this section] for instructions.
+
+== Deploying a Sync Gateway Cluster
+
+The Sync Gateway nodes in a cluster are homogeneous in configuration with the exception of importer nodes.
```
rajagp:
```suggestion
The Sync Gateway nodes in a cluster have a homogeneous configuration with the exception of import node.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+
+Importer node:: Under xref:shared-bucket-access.adoc[convergence/shared bucket access], it is recommended that one node handles document import processing.
```
rajagp:
```suggestion
Importer node:: Under xref:shared-bucket-access.adoc[convergence/shared bucket access], it is recommended that one Sync Gateway node in a cluster be configured for handling document import processing.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+For high availability, you can configure more than one node in your cluster to be an import node, though it is strongly discouraged for all Sync Gateway nodes in the cluster to be configured for import processing.
```
rajagp:
```suggestion
For high availability, you can configure more than one Sync Gateway node in your cluster to be the import node, although it is strongly discouraged for multiple Sync Gateway nodes in the cluster to be configured for import processing.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+The configuration of the Sync Gateway import node is slightly different (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
```
rajagp:
```suggestion
The configuration of the Sync Gateway import node is slightly different than the "regular" Sync Gateway nodes (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+//Replicator node:: if you are using inter-cluster replication using sg-replicate then there will be one designated replicator node whose configuration is different than the rest of the nodes.
+
+The following sections cover the steps to deploy regular nodes or import nodes.
```
rajagp:
```suggestion
The following sections cover the steps to deploy regular or import Sync Gateway nodes.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,314 @@
...
+// or a single replicator node.
+
+For example, to deploy a cluster with 4 regular nodes and 1 importer node you would run the steps below 2 times, each time with a different configuration and deployment controller file.
```
rajagp:
```suggestion
You will have to repeat the steps in this section once for the "regular" nodes and once for the "import" node. The difference is the  Sync Gateway configuration file and the Kubernetes Deployment Controller config file.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
* A Couchbase Server RBAC user with application access privileges.
Sync Gateway will connect to the Couchbase Server as this RBAC user.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: Already the case.

jamesnocentini (commented):
Apply suggestions.

modules/ROOT/assets/attachments/kubernetes/sgw-config-import.json
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/c954f03659c5eb8e80792e9d74fce83e8646f021

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: It's a relative link pointing to a resource in another repo which the github preview doesn't support.
But it will work when built and deployed with Antora.

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/9189d889bd4efaabd09115b93e17977d8f9a385d

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: Marking this convo as resolved. Copied link to this comment on new PR.

jamesnocentini:
@rajagp all suggestions applied.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
+= Installing on Kubernetes
```
rajagp: @jamiltz  
```suggestion
= Deploying Sync Gateway cluster
```

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz OK. But the other links seemed to work - curious why we can't use the same approach? Don't have a strong opinion but wondering...

modules/ROOT/pages/installing-kubernetes.adoc
rajagp: @jamiltz OK. Thanks. Perhaps it will work on Antora because when I tried, it took to top level for some reason.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+      "revs_limit": 20
+    }
+  }
+}
+----
+<1> The server key should point to the Couchbase Server cluster.
+This would typically be of the form `http://CB_SERVER_POD.CB_SERVER_SERVICE_NAME:8091`.
+<2> The username/password keys should match what was setup in the <<pre-requisites, pre-requisites>> section when you configured the Sync Gateway RBAC user.
+. You will use a https://kubernetes.io/docs/concepts/configuration/secret/[Kubernetes Secret] to pass the configuration file to Sync Gateway on launch.
+Alternatively, you could use a https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/[Kubernetes configMap] if you are not concerned about the security.
```
rajagp:
```suggestion
Alternatively, you could use a https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/[Kubernetes configMap] if you are not concerned about security.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+You will have to repeat the steps in this section once for the "regular" nodes and once for the "import" node. The difference is the  Sync Gateway configuration file and the Kubernetes Deployment Controller config file.
+
+=== Configuration file
+
+In this section, you will create the Sync Gateway configuration file and store it in a secret location.
+
+. Open the configuration file corresponding to your deployment.
++
+|===
+| Regular Node | Importer Node
```
rajagp: @jamiltz  Lets use "Import Node" everywhere...
```suggestion
| Regular Node | Import Node
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+For high availability, you can configure more than one Sync Gateway node in your cluster to be the import node, although it is strongly discouraged for multiple Sync Gateway nodes in the cluster to be configured for import processing.
+The configuration of the Sync Gateway import node is slightly different than the "regular" Sync Gateway nodes (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
+//Replicator node:: if you are using inter-cluster replication using sg-replicate then there will be one designated replicator node whose configuration is different than the rest of the nodes.
+
+The following sections cover the steps to deploy regular or import Sync Gateway nodes.
+// or a single replicator node.
+
+You will have to repeat the steps in this section once for the "regular" nodes and once for the "import" node. The difference is the  Sync Gateway configuration file and the Kubernetes Deployment Controller config file.
+
+=== Configuration file
```
rajagp: @jamiltz 
```suggestion
=== Sync Gateway Configuration File
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+In this section, you will create the Sync Gateway configuration file and store it in a secret location.
```
rajagp: @jamiltz 
```suggestion
In this section, you will create a [Kubernetes secret](url
https://kubernetes.io/docs/concepts/configuration/secret/) from a new or existing Sync Gateway configuration.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+. Open the configuration file corresponding to your deployment.
```
rajagp: @jamiltz 
```suggestion
Open the Sync Gateway Configuration file corresponding to your deployment. If you don't have a config file. you may also use our sample config file shown below as a starting point.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+      "password": "password",
+      "users": { "GUEST": { "disabled": false, "admin_channels": ["*"] } },
+      "allow_conflicts": false,
+      "revs_limit": 20
+    }
+  }
+}
+----
+<1> The server key should point to the Couchbase Server cluster.
+This would typically be of the form `http://CB_SERVER_POD.CB_SERVER_SERVICE_NAME:8091`.
```
rajagp: @mattcarabine  Can you confirm that this is accurate representation of path to server pod. Is this service name or cluster name that follows pod name?

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+      "username": "admin", // <2>
+      "password": "password",
+      "users": { "GUEST": { "disabled": false, "admin_channels": ["*"] } },
+      "allow_conflicts": false,
+      "revs_limit": 20
+    }
+  }
+}
+----
+<1> The server key should point to the Couchbase Server cluster.
```
rajagp: @jamiltz 
```suggestion
<1> The server key should point to any pod in the  Couchbase Server cluster.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+This would typically be of the form `http://CB_SERVER_POD.CB_SERVER_SERVICE_NAME:8091`.
+<2> The username/password keys should match what was setup in the <<pre-requisites, pre-requisites>> section when you configured the Sync Gateway RBAC user.
+. You will use a https://kubernetes.io/docs/concepts/configuration/secret/[Kubernetes Secret] to pass the configuration file to Sync Gateway on launch.
+Alternatively, you could use a https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/[Kubernetes configMap] if you are not concerned about the security.
+However, since the Sync Gateway contains sensitive information, it is recommended that you create a secret from the config file and pass that to the Sync Gateway.
+Run the following command to create a secret called "sgw-config" or "sgw-config-import" with the corresponding Sync Gateway configuration file.
```
rajagp: @jamiltz Can we have two separate entries please - one for regular and one for import (like a tab?). I would expect people wanting to just  copy and paste the command and run it without editing (as simple as it may be , its possible that folks will get it wrong)

rajagp: @jamiltz 
```suggestion
Run the following command to create a secret called "sgw-config" or "sgw-config-import" corresponding  to regular and import versions of Sync Gateway configuration file respectively
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
++
+[source,console]
+----
+kubectl create secret generic [sgw-config|sgw-config-import] --from-file [configuration-file]
+----
++
+If successful, you will see the following.
```
rajagp: @jamiltz  Again, to be consistent with command, can we have a tab that shows the output. Realize this isn't a big deal but want to keep it as straightforward as possible

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+In this section you will deploy the Sync Gateway cluster with the configuration file that you created above.
+
+You will be using a https://kubernetes.io/docs/concepts/workloads/controllers/deployment/[Kubernetes deployment controller].
+A deployment controller allows you to define the number of Sync Gateway replicas and other parameters.
+
+. Open the controller file corresponding to your deployment.
++
+|===
+| Regular Node | Importer Node
```
rajagp: @jamiltz 
```suggestion
| Regular Node | Import Node
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+
+Importer node:: Under xref:shared-bucket-access.adoc[convergence/shared bucket access], it is recommended that one Sync Gateway node in a cluster be configured for handling document import processing.
+For high availability, you can configure more than one Sync Gateway node in your cluster to be the import node, although it is strongly discouraged for multiple Sync Gateway nodes in the cluster to be configured for import processing.
+The configuration of the Sync Gateway import node is slightly different than the "regular" Sync Gateway nodes (see xref:config-properties.adoc#databases-foo_db-import_docs[databases.$db.import_docs]).
+//Replicator node:: if you are using inter-cluster replication using sg-replicate then there will be one designated replicator node whose configuration is different than the rest of the nodes.
+
+The following sections cover the steps to deploy regular or import Sync Gateway nodes.
+// or a single replicator node.
+
+You will have to repeat the steps in this section once for the "regular" nodes and once for the "import" node. The difference is the  Sync Gateway configuration file and the Kubernetes Deployment Controller config file.
```
rajagp: @jamiltz 
```suggestion
You will have to repeat the steps in this section once for the "regular" or non-import nodes and once for the "import" node. The difference is the  Sync Gateway configuration file and the Kubernetes Deployment Controller config file.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+            limits:
+              cpu: "2" # <7>
+      volumes: # <8>
+        - name: sgw-config-volume
+          secret:
+            secretName: sgw-config
+----
+<1> `metadata.name`: The name of the deployment is "sync-gateway".
+<2> `spec.replicas`: 2 Sync Gateway replicas that are deployed at most.
+* For import node deployment, this *must* be 1 because there can be at most 1 Sync Gateway replicator in a cluster.
```
rajagp:
```suggestion
* For import node deployment, this *is recommended* to be 1.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+If high availability is important, you probably want at least 2 Sync Gateway import nodes.
```
rajagp: @jamiltz 
```suggestion
If high availability is important, you may want at least 2 Sync Gateway import nodes. However, given the overhead of redundant import docs processing by multiple nodes, this is not typically recommended.
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+* For replicator node deployment, this *must* be 1 because there can be at most 1 Sync Gateway import node in a cluster.
```
rajagp: @jamiltz We dont take about replicator nodes anywhere. Once we are done with first version, we can add replicator node support =
```suggestion
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+<3> `containers[].image`: Points to the docker image for Sync Gateway.
+<4> `containers[].args`: Points to the Sync Gateway configuration file named "sgw-config-working.json" which is mounted at the path specified via the `volumeMounts` config.
```
rajagp: @jamiltz 
```suggestion
<4> `containers[].args`: Points to the Sync Gateway configuration file named "sgw-config-working.json" which is mounted at the path specified via the `volumeMounts` config.
* If you are using the sample config files, this would be "sgw-import-config-working.json" for the import node
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+If you do not specify one, the Sync Gateway could spawn as many processes as CPU cores and potentially use up all CPU resources.
+You can learn more about CPU resource assignment https://kubernetes.io/docs/tasks/configure-pod-container/assign-cpu-resource/#if-you-do-not-specify-a-cpu-limit[here].
+We recommend a value of 2 but you should use what is suited for your environment.
+<8> `volumes`: Specifies what to mount.
+In our case, the "secret" with name "sgw-config" corresponding to the Sync Gateway configuration that was created in the previous step is mounted.
+Learn more about Kubernetes volumes https://kubernetes.io/docs/concepts/storage/volumes/[here].
+. Deploy the Sync Gateway cluster from the specified deployment controller file.
++
+[source,console]
+----
```
rajagp: @jamiltz  Same as before. Please include separate tabs for import and regular with exact command that users can copy and paste

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+kubectl create --from-file [deployment-controller-file]
+----
+If successful, you will see the following.
++
+[source,console]
+----
```
rajagp: @jamiltz  Same as before. For consistency, please include separate tabs for import and regular with corresponding response.

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+. You can check the status of the deployment with the following command until all the pods corresponding to the Sync Gateway are in the "Ready" state and the status is "Running".
++
+[source,console]
+----
+kubectl get pods --watch
+----
+The `--watch` option is optional but you use it to be asynchronously notified of  updates to status of the pods instead of having to repeatedly run the command.
++
+If successful, you will see a listing of the Sync Gateway pods that were deployed.
+In the sample output below, we have Couchbase Server and Sync Gateway pods running on the same namespace.
```
rajagp: @jamiltz 
```suggestion
In the sample output below, we have Couchbase Server and Sync Gateway pods running in the same [namespace](https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/).
```

modules/ROOT/pages/installing-kubernetes.adoc
```diff
@@ -0,0 +1,320 @@
...
+Port:                     <unset>  4984/TCP
+TargetPort:               4984/TCP
+NodePort:                 <unset>  32397/TCP
+Endpoints:                10.0.0.34:4984,10.0.0.35:4984
+Session Affinity:         None
+External Traffic Policy:  Cluster
+Events:
+----
+Notice the "endpoints" field and confirm that it corresponds to the Sync Gateway nodes.
+In this example, we have 2 Sync Gateway nodes (two regular nodes).
```
rajagp: @jamiltz 
```suggestion
In this example, we have 2 Sync Gateway nodes .
```

modules/ROOT/pages/managing-kubernetes.adoc
```diff
@@ -0,0 +1,28 @@
...
+[source,console]
+----
+kubectl rollout status deployment sync-gateway
+----
+The deployment controller will ensure that at least one replica is available to serve the incoming requests while the other replicas are being upgraded.
+The load balancer will automatically redirect all requests to the remaining nodes.
++
+*Optional:* If you want to rollback, you can do that with `kubectl rollout undo  deployment sync-gateway`.
+
+== Upgrading the Kubernetes Cluster
```
rajagp: @jamiltz Is this referring CBS cluster upgrade?  Unclear ....

modules/ROOT/pages/managing-kubernetes.adoc
```diff
@@ -0,0 +1,28 @@
...
+. You can verify the status of the rollout as follows
++
+[source,console]
+----
+kubectl rollout status deployment sync-gateway
+----
+The deployment controller will ensure that at least one replica is available to serve the incoming requests while the other replicas are being upgraded.
+The load balancer will automatically redirect all requests to the remaining nodes.
++
+*Optional:* If you want to rollback, you can do that with `kubectl rollout undo  deployment sync-gateway`.
```
rajagp: @jamiltz 
```suggestion
+ If the upgrade fails for any reason , you can rollback or undo the rollout with the following command 
 `kubectl rollout undo  deployment sync-gateway`.
```

rajagp:
@jamiltz  I am done with my review of this PR. Includes a question for confirmation for @mattcarabine

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
In this section, you will create a https://kubernetes.io/docs/concepts/configuration/secret/[Kubernetes secret] from a new or existing Sync Gateway configuration.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
. Open the Sync Gateway configuration file corresponding to your deployment.
If you don't have a configuration file you may also use our sample configuration shown below as a starting point.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
If high availability is important, you may want at least 2 Sync Gateway import nodes.
However, given the overhead of redundant import docs processing by multiple nodes, this is not typically recommended.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
In the sample output below, we have Couchbase Server and Sync Gateway pods running in the same https://kubernetes.io/docs/tasks/administer-cluster/namespaces-walkthrough/[namespace].
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
In this example, we have 2 Sync Gateway nodes.
```

modules/ROOT/pages/managing-kubernetes.adoc
jamesnocentini:
```suggestion
If the upgrade fails for any reason, you can rollback or undo the rollout with the following command.
+
[source,console]
----
kubectl rollout undo  deployment sync-gateway
----
```

modules/ROOT/pages/managing-kubernetes.adoc
jamesnocentini: Referring to the Kubernetes Cluster upgrade.
In response to Eric's feedback https://github.com/couchbase/docs-sync-gateway/pull/35#pullrequestreview-194797246

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
//* For replicator node deployment, this *must* be 1 because there can be at most 1 Sync Gateway import node in a cluster.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini:
```suggestion
<4> `containers[].args`: Points to the Sync Gateway configuration file named "sgw-config-working.json" which is mounted at the path specified via the `volumeMounts` config.
If you are using the sample config files, this would be "sgw-config-import.json" for the import node.
```

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/a230695a63497984aad1f70dcec4f393d55f0804

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/a230695a63497984aad1f70dcec4f393d55f0804

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/a230695a63497984aad1f70dcec4f393d55f0804

modules/ROOT/pages/installing-kubernetes.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/a230695a63497984aad1f70dcec4f393d55f0804

jamesnocentini:
@rajagp all suggestions applied.

modules/ROOT/pages/installing-kubernetes.adoc
mattcarabine: It depends what namespace the CB cluster is in, the full form would be:

```
$CB_SERVER_POD.couchbase.$NAMESPACE.srv:8091
```

Only downside is that the pod you're bootstrapping off could go down and be replaced (meaning subsequent Sync Gateway pods would fail to bootstrap).

The operator does automatically create SRV records which the SDKs can exploit to connect, this is a single DNS entry that provides access to all pods - https://docs.couchbase.com/operator/1.1/services.html#couchbase-srv.
But I guess Sync Gateway doesn't support SRV bootstrapping (the SDKs only added support for this later down the line), so probably not an option.

I guess good solution here would just be to call out this potential pitfall in the docs (along with the DNS format).

rajagp: @mattcarabine  Thats not very different from non-containerized env right? Even today, SGW is configured points to one of CBS nodes in cluster. But makes sense to clarify that. 
That said, the full form seems different than what was configured on SGW(http://cb-example-0000.cb-example:8091) . Let me verify with the URL that you provided above

rajagp: @mattcarabine $`CB_SERVER_POD.couchbase.$NAMESPACE.srv:8091` did not work (the SGW failed to connect to the bucket). I had to use `$CB_SERVER_POD.$CB_CLUSTER_NAME:8091`.  I used standard server side configuration. What did I miss?

rajagp:
@jamiltz  Can you merge and stage this PR some place so I can do another pass before we decide to publish  ? It is getting very hard to review the changes to the changes :-).  I would also like to share it with someone in field team so they can try it out 

Also - can you please include a bullet in Prerequisites section to download the config files (.zip somewhere)

jamesnocentini:
@rajagp merging the PR would automatically publish the content on docs.couchbase.com.
And we don't have a functionality to build previews of PRs.
I see two options:

1. Merging in `release/2.1` and make quick follow up changes.
2. Merging in `release/2.5` which will make it available on docs-staging.couchbase.com but not docs.couchbase.com. Then cherry-pick from release/2.5 to release/2.1 when ready.

jamesnocentini:
> Also - can you please include a bullet in Prerequisites section to download the config files (.zip somewhere)

I think we already make that clear in the step by step instructions https://github.com/couchbase/docs-sync-gateway/pull/35/files#diff-c78950000fc7582540fb6e1d9dd04f27R33 (i.e users don't need to download the configs prior to starting because there is no `.zip`. Files are downloaded as the user progresses through the instructions. Suggest to publish as is now and can look into doing a `.zip` if the current approach doesn't work that well).

rajagp:
@jamiltz Then lets do option 2 and have it on docs-staging.couchbase.com.

rajagp:
@jamiltz  I'll look to the staged doc to check the instructions to download the config. Its quite hard to comprehend the flow in its current form . I prefer a .zip file in addition to having it in step-by-step instructions . So users don't have to go through every section to identify the files and its also in one place so its easy to locate.  That said, I may very well change my mind once I take a look at the cleaned up version...

jamesnocentini:
OK

jamesnocentini:
@rajagp https://docs-staging.couchbase.com/sync-gateway/2.5/kubernetes/deploy-cluster.html

rajagp:
@jamiltz  Looks good. 3 minor tweaks. 
- missed removing Instances of [importer](https://docs-staging.couchbase.com/sync-gateway/2.5/kubernetes/deploy-cluster.html#deploying-a-sync-gateway-cluster) 
- The configuration of the Sync Gateway import node is slightly different than the "regular" Sync Gateway nodes to 
The configuration of the Sync Gateway import node is slightly different than the **"regular” or “non-import”** Sync Gateway nodes 

- Also can be add a note to end of[ Deployment controller  section](https://docs-staging.couchbase.com/sync-gateway/2.5/kubernetes/deploy-cluster.html#deployment-controller) to indicate "Make sure that you have sufficient CPU resources on the node on which the pods are being deployed. Failure to do so will result in an "Insufficient resource" exception when attempting to deploy the pods" 


I think this is good to go.
 Let me get back to you on when we can publish this (likely on Wednesday after I've run this by some folks in field)

jamesnocentini:
@rajagp Changes commit https://github.com/couchbase/docs-sync-gateway/commit/d8e38e53c1cfb062e944283fa5644049c70d22b4

> Let me get back to you on when we can publish this (likely on Wednesday after I've run this by some folks in field)

OK

rajagp:
@jamiltz  : 
Please update the following-
1) Update sgw-config-working.json (also the corresponding content on page)
 `"server": "cb-example-0000.cb-example.default.svc:8091",`
2) Update sgw-import-config-working.json (also the corresponding content on page)
 `"server": "cb-example-0000.cb-example.default.svc:8091",`
3) Update the text which discusses the server URL and state that it should be of the form "CB_SERVER_POD.CB_SERVER_SERVICE_NAME.NAMESPACE.svc" 


Once these changes are made, this is ready to be published

jamesnocentini:
@rajagp published! https://docs.couchbase.com/sync-gateway/2.1/kubernetes/deploy-cluster.html

rajagp:
@jamiltz  Yay!!! thanks

PR #37, opened 2019-01-18
DOC-4435

jamesnocentini:

### DIFF
--- a/modules/ROOT/pages/managing-tombstones.adoc
+++ b/modules/ROOT/pages/managing-tombstones.adoc
@@ -69,11 +69,17 @@ The table below describes those differences.
 |Persisted on the document
 |Persisted as system extended attributes (`xattr._sync`).
 
-2+a|*Additional properties on a tombstone*
+2+a|*Additional user properties on a tombstone*
 |Persisted on the document
 |Not persisted.
 
 NOTE: Setting additional properties on tombstones may appear to be successful through Couchbase Lite SDKs or the Sync Gateway REST APIs but those properties will not be persisted to Couchbase Server.
+
+2+a|*Additional system properties on a tombstone*
+|Persisted on the document
+|Not persisted.
+
+NOTE: Setting additional system properties (i.e `_exp`) on tombstones may appear to be successful through Couchbase Lite SDKs or the Sync Gateway REST APIs but those properties will not be persisted to Couchbase Server (and therefore not synchronized to clients).
 |===
 
 Document operations also have a different impact on tombstones when Shared Bucket Access is enabled/disabled.
--- a/modules/ROOT/pages/managing-tombstones.adoc
+++ b/modules/ROOT/pages/managing-tombstones.adoc
@@ -71,15 +71,11 @@ The table below describes those differences.
 
 2+a|*Additional user properties on a tombstone*
 |Persisted on the document
-|Not persisted.
-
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/managing-tombstones.adoc
```diff
@@ -69,11 +69,17 @@ The table below describes those differences.
...
 |Persisted on the document
 |Not persisted.
 
 NOTE: Setting additional properties on tombstones may appear to be successful through Couchbase Lite SDKs or the Sync Gateway REST APIs but those properties will not be persisted to Couchbase Server.
+
+2+a|*Additional system properties on a tombstone*
+|Persisted on the document
+|Not persisted.
+
+NOTE: Setting additional system properties (i.e `_exp`) on tombstones may appear to be successful through Couchbase Lite SDKs or the Sync Gateway REST APIs but those properties will not be persisted to Couchbase Server (and therefore not synchronized to clients).
```
jamesnocentini: @adamcfraser this addition may be incorrect? (cc @rajagp)

rajagp: @jamiltz  What confirmation are you looking for ?

jamesnocentini: Per the last comment on JIRA, we wanted to clarify the affect of setting system props (prefixed with `_`) on tombstones.
This change adds this case, but I need a review to make sure what is being added here is correct.

rajagp: OK.  From your question, it seemed like you were thinking that this 
"may be incorrect" :-) . Looks right to me but @adamcfraser  can confirm

jamesnocentini: Yes I assumed the behaviour for system properties to be the same as for user properties. That's a blind assumption which rarely turn out to be correct :)

jamesnocentini: ^^^ the line above (number 82) needs a thumbs up (@bbrks @adamcfraser )

adamcfraser: It's a little unclear as written, because we don't replicate _exp to clients in any situation (tombstones or no).  You could just simplify and say that the document expiry won't be affected by setting _exp on a tombstone (at least for the SG case)

jamesnocentini: I see. I think the explanation is more complex than what it needs to be..

Isn't that just a no-op? In which case I would prefer removing that NOTE altogether and have the following only:

```suggestion
|Not persisted (https://en.wikipedia.org/wiki/NOP_(code)[no-op]).
```

adamcfraser:
I don't even think the wikipedia link is necessary - I agree that 'not persisted' is probably sufficient.

modules/ROOT/pages/managing-tombstones.adoc
jamiltz:
```suggestion
NOTE: Setting additional xref:data-modeling.adoc#property-prefix[system properties] on tombstones may appear to be successful through Couchbase Lite SDKs or the Sync Gateway REST APIs but they are not persisted to Couchbase Server.
```

jamesnocentini:
OK thanks for the feedback! Will pull it out.

PR #46, opened 2019-02-19
DOC-4303

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4303

### DIFF
--- a/modules/ROOT/pages/running-replications.adoc
+++ b/modules/ROOT/pages/running-replications.adoc
@@ -31,7 +31,7 @@ However, SG Replicate was designed specifically for a Couchbase Mobile deploymen
 * In deployments with multiple Sync Gateway nodes, only _one_ of the Sync Gateways should be configured for replications.
 If multiple Sync Gateways are configured for replications, it could substantially increase the amount of duplicate work, and therefore should be avoided.
 The limitation is that the system is not guaranteed to be Highly Available: if the Sync Gateway that is chosen to drive the replication goes down or is otherwise removed from the system, then the replications will stop.
-* When running two Sync Gateway clusters with `allow_conflicts: false`, cross-cluster document conflicts will result in that document no longer being replicated.
+* Replication between Sync Gateway databases doesn't support automatic conflict resolution even when the no-conflicts mode is enabled . When running two Sync Gateway clusters with `allow_conflicts: false`, cross-cluster document conflicts will result. Deployments  must implement custom conflict resolver in an external app as specified [here](https://docs.couchbase.com/sync-gateway/2.1/resolving-conflicts.html).
 To avoid this, the application must ensure concurrent, cross-cluster updates are not made to a given document.
 
 == Running replications via the REST API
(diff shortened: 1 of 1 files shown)


jamesnocentini:
Remove unnecessary sentence. 
This limitation doesn't exclude conflict resolution to be used in such a deployment (in this case conflicts are resolved on CBL).

@adamcfraser @rajagp this PR is waiting for your review/approval.

modules/ROOT/pages/running-replications.adoc
```diff
@@ -31,8 +31,7 @@ However, SG Replicate was designed specifically for a Couchbase Mobile deploymen
 * In deployments with multiple Sync Gateway nodes, only _one_ of the Sync Gateways should be configured for replications.
 If multiple Sync Gateways are configured for replications, it could substantially increase the amount of duplicate work, and therefore should be avoided.
 The limitation is that the system is not guaranteed to be Highly Available: if the Sync Gateway that is chosen to drive the replication goes down or is otherwise removed from the system, then the replications will stop.
-* Replication between Sync Gateway databases doesn't support automatic conflict resolution when the no-conflicts mode is enabled (i.e xref:config-properties.adoc#databases-foo_db-allow_conflicts["allow_conflicts": false]).
-When running two Sync Gateway clusters with `allow_conflicts: false`, cross-cluster document conflicts will result in that document no longer being replicated.
+* When running two Sync Gateway clusters with `allow_conflicts: false`, cross-cluster document conflicts will result in that document no longer being replicated.
```
rajagp:
```suggestion
* Replication between Sync Gateway databases doesn't support automatic conflict resolution even when the no-conflicts mode is enabled . When running two Sync Gateway clusters with `allow_conflicts: false`, cross-cluster document conflicts will result. Deployments  must implement custom conflict resolver in an external app as specified [here](https://docs.couchbase.com/sync-gateway/2.1/resolving-conflicts.html).
```

Also , double check w/ @adamcfraser  on what happens if custom conflict resolver is not specified under each of these cases and please clarify in the doc
- allow-conflicts = false . Does conflicting doc get rejected?
- allow-conflicts = true.  Conflict will be created .

rajagp: Also , double check w/ @adamcfraser  on what happens if custom conflict resolver is not specified under each of these cases and please clarify in the doc
- allow-conflicts = false . Does conflicting doc get rejected?
- allow-conflicts = true.  Conflict will be created .

jamesnocentini: @rajagp I don't understand the question. My understanding is `allow_conflicts: false` always rejects conflicting revisions ([docs](https://docs.couchbase.com/sync-gateway/2.1/config-properties.html#databases-foo_db-allow_conflicts)). Can you elaborate?

adamcfraser: That's correct - allow_conflicts:false will not allow conflicts to be created (write attempts will return a 409 error).  Perhaps the original version could be modified to state 'will result in that document no longer being replicated until the conflict is resolved.', and then potentially add some of the information @rajagp has provided?

PR #52, opened 2019-03-12
DOC 4055

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4055

### DIFF
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -73,6 +73,18 @@ properties:
             description: |
               Configures how long Sync Gateway should wait for a bucket operation to complete before timing out and trying again. The value can be increased in scenarios where there is a heavy load on Couchbase Server and operations are likely to take more than 2.5 seconds to complete. The default value is 2500 milliseconds.
             default: 2500
+          delta_sync:
+            type: object
+            description: Delta sync configuration.
+            properties:
+              enabled:
+                type: boolean
+                description: To enable delta sync during replication with this database.
+                default: 'true'
+              rev_max_age_seconds:
+                type: integer
+                description: How long to keep deltas to old revisions for (in seconds).
+                default: 86400
           import_docs:
             type: boolean
             description: |
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -75,15 +75,50 @@ properties:
             default: 2500
           delta_sync:
             type: object
-            description: Delta sync configuration.
+            description: |
+              *NOTE:* Delta Sync Feature is an Enterprise Edition feature on Sync Gateway and Couchbase Lite.
+
+              Delta sync configuration.
+
+              Delta Sync is the ability to replicate only parts of the Couchbase mobile document that have changed.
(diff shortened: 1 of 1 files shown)


modules/ROOT/assets/attachments/sg.yaml
```diff
@@ -73,6 +73,61 @@ properties:
             description: |
               Configures how long Sync Gateway should wait for a bucket operation to complete before timing out and trying again. The value can be increased in scenarios where there is a heavy load on Couchbase Server and operations are likely to take more than 2.5 seconds to complete. The default value is 2500 milliseconds.
             default: 2500
+          delta_sync:
+            type: object
+            description: |
+              *NOTE:* Delta Sync is an Enterprise Edition feature on Sync Gateway and Couchbase Lite.
+
+              Delta Sync is the ability to replicate only parts of the Couchbase mobile document that have changed. This can result in significant savings in bandwidth consumption as well as throughput improvements, especially when network bandwidth is typically constrained.
```
bbrks, line 81: I'm wondering if we should call out the additional bucket storage requirements when enabling delta sync. These can be tuned with the `rev_max_age_seconds` config option, as described further below, but is not mentioned up here in the general description of the feature.

modules/ROOT/assets/attachments/sg.yaml
```diff
@@ -73,6 +73,61 @@ properties:
...
+              - Delta sync is disabled for Couchbase Lite database replicas.
+            properties:
+              enabled:
+                type: boolean
+                description: Set this property to "true" to enable delta sync.
+                default: 'false'
+              rev_max_age_seconds:
+                type: integer
+                description: |
+                  On write, revision body is backed up in bucket and retained for `rev_max_age_seconds`.
```
bbrks: This incurs additional bucket storage requirements, which depend on the size of documents, and the frequency of updates.

This can be estimated with the following: `(doc_size * updates_per_day * 86400) / rev_max_age_seconds`

bbrks (changes requested):
Let me know what you think about the comments left.

modules/ROOT/assets/attachments/sg.yaml
jamesnocentini: Good to know that the value `rev_max_age_seconds` incurs additional bucket storage requirement on CBS. Yes it's something we should document.

I will also add an example scenario:
> So with `rev_max_age_seconds`'s default value, an average doc size of 4K and 100 writes/day; enabling delta sync would take up an additional 400K on CBS `(4*100*86400)/86400`.

modules/ROOT/assets/attachments/sg.yaml
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/b316f29a3da260cf3599cb71f0474196ac6047ba

modules/ROOT/assets/attachments/sg.yaml
jamesnocentini: @bbrks Done in https://github.com/couchbase/docs-sync-gateway/commit/f3c49a79ea1b201dce1b5dbbb6357d103e20f080.
I will merge this to get 2.5 out the door but I would appreciate your re-review and if it needs further amendments.

PR #53, opened 2019-03-22
Small logging fixes

bbrks:
- Added missing escapes for Windows path backslashes

### DIFF
--- a/modules/ROOT/pages/logging.adoc
+++ b/modules/ROOT/pages/logging.adoc
@@ -202,7 +202,7 @@ On MS Windows `logFilePath` supports the following path formats.
 [source,javascript]
 ----
 "C:/var/tmp/sglogfile.log"
-`C:\var\tmp\sglogfile.log`
+`C:\\var\\tmp\\sglogfile.log`
 `/var/tmp/sglogfile.log`
 "/var/tmp/sglogfile.log"
 ----


modules/ROOT/pages/logging.adoc
```diff
@@ -202,7 +202,7 @@ On MS Windows `logFilePath` supports the following path formats.
 [source,javascript]
 ----
 "C:/var/tmp/sglogfile.log"
-`C:\var\tmp\sglogfile.log`
+`C:\\var\\tmp\\sglogfile.log`
 `/var/tmp/sglogfile.log`
 "/var/tmp/sglogfile.log"
```
bbrks, line 207: @jamiltz I'm not sure if this is a valid Windows log file path or not. I can see how omitting the leading `/` would make it relative to SG, but can't say without testing.

jamesnocentini: It looks like Windows absolute paths can start with a leading `\` https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

Does that clarify? (from what I understand `\\` == `/` on windows?)

modules/ROOT/pages/logging.adoc
```diff
@@ -202,7 +202,7 @@ On MS Windows `logFilePath` supports the following path formats.
 [source,javascript]
 ----
 "C:/var/tmp/sglogfile.log"
-`C:\var\tmp\sglogfile.log`
+`C:\\var\\tmp\\sglogfile.log`
```
jamesnocentini, line 205: A single backslash seems valid looking at https://docs.microsoft.com/en-us/dotnet/standard/io/file-path-formats

Am I missing something on this page?

modules/ROOT/pages/logging.adoc
bbrks: Oops, I'd not noticed the backticks around the 2nd line - You only need to escape backslashes in double quotes in JSON. This PR can be closed! :) 

The leading slash makes sense to me after reading that doc - Go will translate forward slashes to Windows-specific backslashes - so should work fine!

bbrks:
Closing - see comment

modules/ROOT/pages/logging.adoc
jamesnocentini: 👍

PR #56, opened 2019-04-05
Fixed commands that use -- from-file to use -f

rajagp:
https://issues.couchbase.com/browse/DOC-5040

### DIFF
--- a/modules/ROOT/pages/kubernetes/deploy-cluster.adoc
+++ b/modules/ROOT/pages/kubernetes/deploy-cluster.adoc
@@ -102,7 +102,7 @@ Regular node::
 --
 [source,console]
 ----
-kubectl create secret generic sgw-config --from-file sgw-config.json
+kubectl create secret generic sgw-config -f sgw-config.json
 ----
 --
 
@@ -111,7 +111,7 @@ Import node::
 --
 [source,console]
 ----
-kubectl create secret generic sgw-config-import --from-file sgw-config-import.json
+kubectl create secret generic sgw-config-import -f sgw-config-import.json
 ----
 --
 ====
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/kubernetes/deploy-cluster.adoc
```diff
@@ -102,7 +102,7 @@ Regular node::
 --
 [source,console]
 ----
-kubectl create secret generic sgw-config --from-file sgw-config.json
+kubectl create secret generic sgw-config -f sgw-config.json
```
mattcarabine, line 105: This one is actually right! It creates a secret from a file, it's just the deployment ones that were wrong.

modules/ROOT/pages/kubernetes/deploy-cluster.adoc
```diff
@@ -111,7 +111,7 @@ Import node::
 --
 [source,console]
 ----
-kubectl create secret generic sgw-config-import --from-file sgw-config-import.json
+kubectl create secret generic sgw-config-import -f sgw-config-import.json
```
mattcarabine, line 114: This one is actually right! It creates a secret from a file, it's just the deployment ones that were wrong.

mattcarabine (changes requested):

modules/ROOT/pages/kubernetes/deploy-cluster.adoc
jamesnocentini: Back when we were working on this, I thought `--from-file` would work on all commands and be preferable to `-f` as it is more explicit. But given it's not the case let's use `-f` everywhere for consistency.

PR #59, opened 2019-04-23
DOC-4836

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4836

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -42,4 +42,5 @@
 
 .Release Notes & Compatibility
 * xref:release-notes.adoc[Release Notes]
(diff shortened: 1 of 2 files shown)


eric-schneider:
I think it would be helpful to provide a definition of what 'deprecated' means.

jamesnocentini:
Good point. I would like to find a good way to channel that across all docs (so I can re-use that definition in other components). I have been keeping a wip of a [mobile glossary](https://docs.couchbase.com/sync-gateway/2.1/glossary.html) but would ideally prefer to merge it to an all-docs glossary. One idea could be to have a new repository called `docs-glossary` that would keep the `glossary` component with a single version (branch). Or another option, I haven't thought about it much.

PR #61, opened 2019-04-24
Update 2.5 Known Issues

JFlath:
(Not sure if I've messed up any new lines there - seems like it always had that line break)


jamesnocentini:
Hmm not sure either.

PR #66, opened 2019-05-13
DOC-2920

jamesnocentini:


modules/ROOT/pages/stats-monitoring.adoc
```diff
@@ -196,6 +196,10 @@ This metric can be used to calculate the time between seeing a change on the DCP
 
 DCP cache latency = `dcp_caching_time` / `dcp_caching_count`
 
+===== max_pending
+
+Max number of documents waiting on a missing earlier sequence number.
```
jamesnocentini: @adamcfraser @bbrks is that a stat we expect users to use often?

If that is the case, should it contain a bit more information about the underlying mechanism being tracked by this stat?

adamcfraser: We don't expect users to use often - this is probably primarily useful to support.  However, could reword as "High watermark for number of documents buffered during feed processing, waiting on a missing earlier sequence."

jamesnocentini: @adamcfraser thanks

PR #67, opened 2019-05-13
DOC-2920

jamesnocentini:

### DIFF
--- a/modules/ROOT/pages/stats-monitoring.adoc
+++ b/modules/ROOT/pages/stats-monitoring.adoc
@@ -191,7 +191,7 @@ This is the pre-compressed size.
 
 ===== attachment_pull_count
 
-Number of attachments pulled per second.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/stats-monitoring.adoc
```diff
@@ -184,6 +184,15 @@ Revision cache misses.
...
 ==== cbl_replication_pull
 
+===== attachment_pull_bytes
+
+Average size of attachments pulled.
+This is the pre-compressed size.
+
+===== attachment_pull_count
+
+Number of attachments pulled per second.
```
bbrks: This stat is just a cumulative count of attachments pulled. I think we were relying on a UI-specific thing to work out the rate.

```suggestion
Number of attachments pulled.
```

bbrks (changes requested):

PR #69, opened 2019-05-14
DOC-5227: Fix non-import config.

JFlath:
I honestly give up with editing in GitHub and new lines at the end of files...


jamesnocentini:
@JFlath same here!

PR #76, opened 2019-05-24
DOC-5288: Fix incorrect logging config

JFlath:
Also enables console logging and uses default logging settings

### DIFF
--- a/modules/ROOT/assets/attachments/kubernetes/sgw-config-import.json
+++ b/modules/ROOT/assets/attachments/kubernetes/sgw-config-import.json
@@ -1,32 +1,10 @@
 {
   "logging": {
+    "log_file_path": "/var/tmp/sglogs",
     "console": {
-      "log_file_path": "/var/tmp/sglogs",
-      "console": {
-        "enabled": false,
-        "log_level": "debug",
-        "log_keys": ["*"]
-      },
-      "error": {
-        "enabled": true,
-        "rotation": {
-          "max_size": 20,
-          "max_age": 180
-        }
-      },
-      "warn": {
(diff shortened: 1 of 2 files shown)


modules/ROOT/assets/attachments/kubernetes/sgw-config-import.json
```diff
@@ -1,32 +1,10 @@
...
-      },
-      "info": {
-        "enabled": false
-      },
-      "debug": {
-        "enabled": false
-      }
+      "enabled": true,
+      "log_level": "info",
+      "log_keys": ["*"]
```
bbrks, line 7: Do we want to log `"*"` by default? What about just keeping the default of `"HTTP"`?

I wonder if keeping everything default/omitted, apart from `log_file_path` and `console.enabled` would simplify the config a bit?

bbrks (commented):
What do you think about this?

modules/ROOT/assets/attachments/kubernetes/sgw-config-import.json
JFlath: I'm on the fence - I'm tempted to say `*` as it give a bit more feedback for those trying this for the first time (kinda who the linked page is aimed at). Having the params listed makes it nice and obvious what can be tuned down too...

bbrks: That's fair enough - I guess with upcoming changes, even `"*"` at `INFO` shouldn't be too problematic.

jamesnocentini (approved):

PR #86, opened 2019-06-18
Doc 2920 resource utilization stats

jamesnocentini:
https://issues.couchbase.com/browse/DOC-2920?filter=16548

@adamcfraser @bbrks I'm not sure about the public and admin interface definitions. Would appreciate your approval on that.

The rest should be ok but feel free to comment if you spot anything.


modules/ROOT/pages/stats-monitoring.adoc
```diff
@@ -118,6 +118,90 @@ Global Sync Gateway stats
...
+
+===== pub_net_bytes_sent
+
+Number of bytes sent on the public interface.
+
+Throughput on public interface = `pub_net_bytes_recv` / `pub_net_bytes_sent`
+
+===== system_memory_total
+
+Total memory available on the system.
```
jamesnocentini: @adamcfraser @bbrks is `system_memory_total` in bytes?

bbrks: yes

modules/ROOT/pages/stats-monitoring.adoc
```diff
@@ -118,6 +118,90 @@ Global Sync Gateway stats
 
 Resource utilization stats
 
+===== admin_net_bytes_recv
+
+Number of bytes received on the admin interface.
+
+Throughput on admin interface = `admin_net_bytes_recv` / `admin_net_bytes_sent`
```
bbrks, line 127: This is the total number of bytes sent on your network interface that the SG admin port is bound to, since boot. So it's not only counting traffic through SG - it's all other stuff happening on the same NIC.

The reason we have different admin_ and pub_ stats, is that `127.0.0.1:4985` and `0.0.0.0:4984` are bound to different network interfaces.

bbrks (changes requested):

modules/ROOT/pages/stats-monitoring.adoc
jamesnocentini: https://github.com/couchbase/docs-sync-gateway/commit/052aca737a9fbc8087390c410065aea19b4e7299

modules/ROOT/pages/stats-monitoring.adoc
jamesnocentini: Updated in https://github.com/couchbase/docs-sync-gateway/commit/f238846706d8e3b144ed851d6e27392fc7b2d35e

> So it's not only counting traffic through SG - it's all other stuff happening on the same NIC.

Oh ok, additional note to clarify https://github.com/couchbase/docs-sync-gateway/commit/4c2d9a87bb0bb156b7dfc41214d695ec34999998

adamcfraser (approved):
LGTM

PR #88, opened 2019-06-21
DOC-2920 - hit ratio derived stats

jamesnocentini:

[Row 21 and 22](https://docs.google.com/spreadsheets/d/1aZjCIwwZYrUSaQCK63QUg7_jrbo0jJcsxPoXG2Q4s0o/edit#gid=0) give a definition of channel and rev cache hit ratio.

@adamcfraser @bbrks I am wondering if we should add a link to the config props that play a part in the hit ratio. Can you remind me which one it is?


jamesnocentini:
I've just realised setting a channel cache limit is a cobalt feature, so not relevant to this PR, but will be for cobalt docs. Merging.

PR #89, opened 2019-06-21
DOC-2920 - gsi view/query stats

jamesnocentini:

[Row 87 to 91](https://docs.google.com/spreadsheets/d/1aZjCIwwZYrUSaQCK63QUg7_jrbo0jJcsxPoXG2Q4s0o/edit#gid=0) mention a few stats in the "Underlying Stats" column (also in this PR). However I could not see them in the [sample output](https://gist.github.com/jamiltz/261589296daa6fcb1022fe67520a90d2) I'm using.

@adamcfraser @bbrks Are those returned by default?

### DIFF
--- a/modules/ROOT/pages/stats-monitoring.adoc
+++ b/modules/ROOT/pages/stats-monitoring.adoc
@@ -99,6 +99,17 @@ The response is a JSON object with the following schema.
           "{xref-cb-config}num_access_errors[num_access_errors]": 0,
           "{xref-cb-config}num_docs_rejected[num_docs_rejected]": 0,
           "{xref-cb-config}total_auth_time[total_auth_time]": 0
+        },
+        "??": {
+          "channels_count": 0,
+          "access_count": 0,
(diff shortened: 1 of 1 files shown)


adamcfraser:
The gsi_views stats aren't initialized to zero (like the rest of the stats) - they are currently only initialized on first use.  Your sample output includes a few of them (`access_count`, etc).  

@jamiltz Does that address your question, or are there more details we can provide?

ibsoln:
@rajagp @adamcfraser Don't know what to do with this. Are the fields James included correct or not? Let me know so I can close this off, please.

rajagp:
@adamcfraser  can comment.
@ibsoln  I don't know what the question is about. This [gist](https://gist.github.com/rajagp/77a0965a3971dabae1f30f9d54f02845) is a sample output of our stats (its from 2.5 but the ticket was from that time). You can compare it with what's published on our docs and see if anything is missing

PR #91, opened 2019-06-25
DOC-2929 - sequence values as integer or dictionary

jamesnocentini:
document `since` and `last_seq` values as integer or dictionary

https://issues.couchbase.com/browse/DOC-2929

@adamcfraser @bbrks I'm not sure this is correct, would you mind providing a review? The JIRA issue talks about JSON values, I assume that's a dictionary (`type: object` in the OpenAPI spec)

### DIFF
--- a/modules/ROOT/assets/attachments/sync-gateway-admin.yaml
+++ b/modules/ROOT/assets/attachments/sync-gateway-admin.yaml
@@ -265,7 +265,7 @@ parameters:
   since:
     in: query
     name: since
-    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs can be integer or dictionary types and should be considered opaque; they come from the last_seq property of a prior response.
+    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
     type: integer
     required: false
   style:
@@ -351,7 +351,7 @@ parameters:
           type: string
           default: 'normal'
         since:
-          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs can be integer or dictionary types and should be considered opaque; they come from the last_seq property of a prior response.
+          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
           type: object
         heartbeat:
           description: Default is 0. Interval in milliseconds at which an empty line (CRLF) is written to the response. This helps prevent gateways from deciding the socket is idle and closing it. Only applicable to longpoll or continuous feeds. Overrides any timeout to keep the feed alive indefinitely. Setting to 0 results in no heartbeat.
--- a/modules/ROOT/assets/attachments/sync-gateway-admin.yaml
+++ b/modules/ROOT/assets/attachments/sync-gateway-admin.yaml
@@ -265,7 +265,7 @@ parameters:
   since:
     in: query
     name: since
-    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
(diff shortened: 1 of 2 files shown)


modules/ROOT/assets/attachments/sync-gateway-admin.yaml
```diff
@@ -265,7 +265,7 @@ parameters:
   since:
     in: query
     name: since
-    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
+    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs can be integer or dictionary types and should be considered opaque; they come from the last_seq property of a prior response.
```
adamcfraser: I think it might be best to just leave it as 'Sequence IDs should be considered opaque', and change the type to 'object' everywhere.  Today they can be integer or JSON String - I don't think there's a case where we send dictionary - but I agree it's correct they should be considered an opaque object.

adamcfraser (changes requested):

modules/ROOT/assets/attachments/sync-gateway-admin.yaml
jamesnocentini:
```suggestion
    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
```

modules/ROOT/assets/attachments/sync-gateway-admin.yaml
```diff
@@ -351,8 +351,8 @@ parameters:
           type: string
           default: 'normal'
         since:
-          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
-          type: integer
+          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs can be integer or dictionary types and should be considered opaque; they come from the last_seq property of a prior response.
```
jamesnocentini:
```suggestion
          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
```

modules/ROOT/assets/attachments/sync-gateway-public.yaml
```diff
@@ -266,8 +266,8 @@ parameters:
   since:
     in: query
     name: since
-    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
-    type: integer
+    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs can be integer or dictionary types and should be considered opaque; they come from the last_seq property of a prior response.
```
jamesnocentini:
```suggestion
    description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
```

modules/ROOT/assets/attachments/sync-gateway-public.yaml
```diff
@@ -352,8 +352,8 @@ parameters:
           type: string
           default: 'normal'
         since:
-          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
-          type: integer
+          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs can be integer or dictionary types and should be considered opaque; they come from the last_seq property of a prior response.
```
jamesnocentini:
```suggestion
          description: Starts the results from the change immediately after the given sequence ID. Sequence IDs should be considered opaque; they come from the last_seq property of a prior response.
```

jamesnocentini (commented):
applying changes

PR #94, opened 2019-07-04
DOC-5184 - `max_number` updates

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5184

Adding content per feedback.

@adamcfraser @bbrks Ready for review. Thanks!

### DIFF
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -201,8 +201,6 @@ properties:
                   compact_high_watermark_pct:
                     type: integer
                     description: |
-                      Enterprise Edition.
-
                       High watermark for channel cache eviction (percent).
 
                       When the cache size, determined by `max_number`, reaches the high watermark, the eviction process iterates through the cache to remove inactive channels.
@@ -210,8 +208,6 @@ properties:
                   compact_low_watermark_pct:
                     type: integer
                     description: |
-                      Enterprise Edition.
-
                       Low watermark for channel cache eviction (percent).
 
                       When the cache size, determined by `max_number` returns to a value lower than `compact_low_watermark_pct`, the cache eviction process is stopped.
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -226,12 +226,17 @@ properties:
 
                       The default value for this property is 50000. Along with the default channel `min_length` and `max_length` values, this would result in a memory usage under 1GB.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/assets/attachments/sg.yaml
```diff
@@ -205,25 +205,34 @@ properties:
...
                     type: integer
                     description: |
                       Enterprise Edition.
 
-                      Maximum number of channel caches which will exist at any one point.
+                      Maximum number of channel caches which will exist at any one point. This property is used to determine the cache size (and the associated eviction watermarks `compact_low_watermark_pct`/`compact_high_watermark_pct`).
+
+                      The default value for this property is 50000. Along with the default channel `min_length` and `max_length` values, this would result in a memory usage under 1GB.
+
+                      The `max_number` value should be higher than the estimated number of replicating clients to optimize cache usage and minimize the number of cache misses.
```
adamcfraser: It's not actually the number of replicating clients, since multiple clients could be replicating the same channel(s).  

One option might be to mention the cache hit/miss stats here (cache.chan_cache_hits, cache.chan_cache_misses), and mention that this value can be tuned to optimize cache hits.

modules/ROOT/assets/attachments/sg.yaml
```diff
@@ -205,25 +205,34 @@ properties:
...
 
-                      It should be set to equal the estimated number of replicating clients to avoid the impact of cache misses.
+                      If the cache size grows to reach the high watermark (`compact_high_watermark_pct`), channels with no connected replications will be evicted before channels which are associated with an active pull replication (i.e a blip-based pull replication in Couchbase Lite 2.x, or an active `/{db}/_changes` request in Couchbase Lite 1.x).
 
                       The minimum allowed value is 1.
+                      Setting the value to zero removes the limit to the number of channels being cached.
```
adamcfraser: A few corrections here:
- The minimum allowed value is 100
- It's not possible to set the value to zero.  Users who want to 'remove the limit' would need to set this to an arbitrarily high value - there's no built-in option to revert to 2.5-like behaviour.

adamcfraser (changes requested):
A few changes/clarifications.

jamesnocentini:
@adamcfraser Updated per your review. I could not see `cache.chan_cache_hits` and `cache.chan_cache_misses` in [this spec](https://docs.google.com/document/d/1v60oiWY-M0qExEnWb-bvEodj76ds_qfJXoL3U_9locA/edit?ts=5cc1af33).

Are those props new in colbalt or were added in a previous version?

PR #98, opened 2019-07-22
DOC-4726 - review `register` property

jamesnocentini:
https://issues.couchbase.com/browse/DOC-4726

@bbrks I have a follow-up PR here to have a bit more info for `register` properties. I hope that change is accurate.

One bit missing might be to write down how Sync Gateway works out if a user with that `id_token`/`access_token` already exists in SG or is a new user?

### DIFF
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -562,7 +562,7 @@ properties:
     type: object
     properties:
       register:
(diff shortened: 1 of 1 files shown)


bbrks:
> One bit missing might be to write down how Sync Gateway works out if a user with that id_token/access_token already exists in SG or is a new user?

The Facebook new user creation is based purely on the email returned to us by Facebook.

For Google, there's a username and also an email address that we'll check before creating a new user.

jamesnocentini:
👍

PR #108, opened 2019-10-24
import filter in getting started

jamesnocentini:
@rajagp This PR adds tabs in the [getting started](https://docs.couchbase.com/sync-gateway/2.6/getting-started.html#configure-sync-gateway) with the current config as the default option, but with the addition of a config that contains an import filter. To help readers which have an existing cluster with a large amount of documents. Do you think it would help? (cc @djpongh)

### DIFF
--- a/modules/ROOT/examples/getting-started/sync-gateway-config-import-filter.json
+++ b/modules/ROOT/examples/getting-started/sync-gateway-config-import-filter.json
@@ -1,0 +1,29 @@
+{
+  "log": ["*"],
+  "databases": {
+    "getting-started-db": {
+      "server": "http://localhost:8091",
+      "bucket": "getting-started-bucket",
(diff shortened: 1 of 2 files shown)


rajagp:
@jamiltz  Yes thats reasonable. I think with 2.7, we should probably clarify the default values or link to docs that describe the same

jamesnocentini:
@rajagp Thanks, I'll merge. And have added the note to clarify the default values in 2.7 on my list [on JIRA](https://issues.couchbase.com/browse/DOC-5711?focusedCommentId=363854&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-363854).

jamesnocentini:
The update is published https://docs.couchbase.com/sync-gateway/2.6/getting-started.html#configure-sync-gateway.

PR #109, opened 2019-10-24
DOC-5711 - Write Scaling Enhancements

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5711


jamesnocentini:
@bbrks , @sarathkumarsivan , @adamcfraser 

At some point before the release, this PR will need with a review for accuracy from Engineering. Only one review should be sufficient. I have assigned the corresponding JIRA ticket (in the comment above) to "The One".

jamesnocentini:
Needs review from @rajagp

modules/ROOT/assets/attachments/sg.yaml
```diff
@@ -144,8 +144,24 @@ properties:
             description: |
               This property works in conjunction with the [enable_shared_bucket_access](#databases-foo_db-enable_shared_bucket_access) property. When **shared bucket access** is enabled, setting this property to true (`"import_docs": true`) indicates that this Sync Gateway node should perform import processing.
 
-              `enable_shared_bucket_access` needs to be `true` on all nodes participating in such a configuration where as `import_docs` can only be set to `true` on a single node.
+              Starting in Sync Gateway 2.7, all Sync Gateway nodes can be configured as import nodes. This results in performance benefits as the import process is shared across all Sync Gateway nodes.
+
+              Prior to version 2.7, `import_docs` can only be set to `true` on a single node.
```
ibsoln: @priya Is this still true at 2.7? If enable_shared_bucket_access = true doesn't import_docs default to 'true'? Also won't multiple nodes need it set to true in order enable the position of multiple nodes handling the import processing work load? Setting it false enables workload isolation for those nodes excluded from import processing and dedicated to handling remote clients.

ibsoln:
@RichardSmedley / @rajagp can you please mark as reviewed so I can complete the merge?

ibsoln (approved):

modules/ROOT/pages/index.adoc
```diff
@@ -2,12 +2,13 @@
...
-Sync Gateway 2.6 supports enhancements to Sync Gateway’s cache management (channels, revisions) and supports additional configuration options to tune Sync Gateway cache settings.
-The smarter cache management techniques will allow the Sync Gateway to scale well with growth in the number of channels while the configuration options will give users better control over cache sizing.
+Sync Gateway 2.7 introduces enhancements which improve the write throughput.
+All Sync Gateway nodes can now participate in import processing, where as this was previously constrained to a single node.
+In addition, it guarantees the high availability of the import process and means that all Sync Gateway nodes can be deployed with the same configuration file.
 
-xref:deployment.adoc#channel-and-revision-cache[Read more]
+xref:shared-bucket-access.adoc[Read more]
 
 == Release Notes
```
rajagp, line 13: Please include following 
**Write Throughput Enhancements** 
We have improved the per-node throughput while processing writes that come in from Couchbase Lite clients [replicating](https://docs.couchbase.com/couchbase-lite/2.7/swift.html#starting-a-replication) with the Sync Gateway, as well as writes that are made via the Sync Gateway [REST API](https://docs.couchbase.com/sync-gateway/current/rest-api.html#/document).

**Import Processing Throughput** 
we have also significantly improved the performance of writes coming in from the Couchbase Server that are [imported](https://docs.couchbase.com/sync-gateway/2.6/config-properties.html#databases-foo_db-enable_shared_bucket_access) by the Sync Gateway . In addition to increasing the per-node import throughput, the enhancements also enable to scale linearly with the number of nodes. 

**High Availability of Imports**
In 2.7, import processing is automatically sharded across all the Sync Gateway nodes in a cluster. That way, even if the import node goes down, other nodes in the cluster can pick up the import shard. That way, import processing can continue uninterrupted with the other nodes in the cluster.

rajagp: @ibsoln  Please update the Whats New Page . You will need to add links

PR #110, opened 2019-10-24
add missing quotes

jamesnocentini:
@adamcfraser @bbrks @sarathkumarsivan although this is only a syntax change, could you confirm if the default value of `import_docs` is indeed `false` in 2.6 and will be `true` in 2.7?


rajagp:
@jamiltz It is not clear cut. Value depends on the value of shared bucket access . Please add table as described in https://docs.google.com/document/d/1h9ZRmVJK6l-FTAAqbF7skjsj_n8xZdvs1FUfmNC7htw/edit#heading=h.hp4ybpwfaqd3

And if you need to provide a value in table, specify the value depending on the default value of shared-bucket-access (which I believe is disabled?)

jamesnocentini:
@rajagp Got it. I had not realized the default value changed with `enable_shared_bucket_access`. Thanks Priya.

PR #117, opened 2019-11-15
DOC-5782 - replace swagger spec links

jamesnocentini:
https://issues.couchbase.com/browse/DOC-5782

Changes:

- replace old swagger refs (.json extension) with docs.couchbase.com links (.yaml extension)

Swagger specs are only published as yaml currently.
It looks like it should also work loading from a yaml spec (swagger-js example https://github.com/swagger-api/swagger-js/blob/2.x/test/client.js#L547-L551)

Keeping open to investigate and resolve.

### DIFF
--- a/modules/ROOT/pages/integrating-external-stores.adoc
+++ b/modules/ROOT/pages/integrating-external-stores.adoc
@@ -62,7 +62,7 @@ var api = 'http://localhost:8000/movies'
   , db = 'movies_lister';
 
 var client = new Swagger({
-  url: 'http://developer.couchbase.com/mobile/swagger/sync-gateway-admin/spec.json',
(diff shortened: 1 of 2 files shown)


rajagp (commented):
@jamiltz  Please add a statement "We do not guarantee that the swagger spec will be aligned with the latest version of the REST API. The REST API must be considered as the source of truth and in case of any deviations, the REST API will override the swagger spec. So please consider the spec as a starting point and make any relevant changes as needed to ensure that it is in conformance with the REST API."

jamesnocentini:
@rajagp Ok thanks for the detailed text, added in https://github.com/couchbase/docs-sync-gateway/pull/117/commits/96c02b99f2bd15ef60d9689001207e8a38ae6fde

ibsoln (approved):

PR #142, opened 2019-12-06
terminology improvement

RichardSmedley:
https://tools.ietf.org/id/draft-knodel-terminology-00.html#master-slave


simon-dew (approved):
LGTM

PR #144, opened 2019-12-11
DOC-5709-RelNotes-SG

ibsoln:
https://issues.couchbase.com/browse/DOC-5709
subtask https://issues.couchbase.com/browse/DOC-6701

Added latest SG Release Notes


bmeike (approved):
Are these macros?  The idea is that the wording will always be the same?  Seems like a good idea.
I cannot speak to the accuracy of the content.  You'll need at least one member of the SGW team to do that

PR #156, opened 2020-01-22
DOC-5902-for2.0

ibsoln:
https://issues.couchbase.com/browse/DOC-5902


ibsoln:
This change isn't a straightforward 'apply to previous versions'. I need to manually apply the appropriate changes for the individual releases instead of using git stash apply 'n'.

PR #159, opened 2020-01-22
DOC-5902-Revert-Changes-To-1.5

ibsoln:
Want to back out the past two changes as the first of them contained references to allow_conflicts, which was not introduced unitl 2.x releases.
When I resubmit the amended DOC-5902 changes I will ensure that the YAML file is validated; avoiding the potential for not displayed to configuration page.


RichardSmedley (approved):
LGTM (afaict ;) but did you mean to cut minimum out after line 314?)

PR #161, opened 2020-01-24
Remove test-only fixes from release notes

adamcfraser:


ibsoln (approved):
Thank you @adamcfraser

PR #162, opened 2020-01-27
DOC-6142 - Deprecate command line option -bucket from Sync Gateway

ibsoln:
https://issues.couchbase.com/browse/DOC-6142

Deprecate CLI -bucket option at 2.7


RichardSmedley (approved):
:+1:

PR #176, opened 2020-02-05
DOC-6137 Misleading distribution statement

ibsoln:
https://issues.couchbase.com/browse/DOC-6137

Changes to shared-bucket-access.adoc

### DIFF
--- a/modules/ROOT/pages/shared-bucket-access.adoc
+++ b/modules/ROOT/pages/shared-bucket-access.adoc
@@ -55,7 +55,7 @@ The import process is a key part of mobile convergence.
 It is the means by which Sync Gateway becomes aware of non-Sync Gateway data changes and obtains the mobile metadata it requires to replicate changes.
 Any link:glossary.adoc#non-sync-gateway[non-Sync Gateway change] is eligible for import.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/shared-bucket-access.adoc
```diff
@@ -55,7 +55,7 @@ The import process is a key part of mobile convergence.
 It is the means by which Sync Gateway becomes aware of non-Sync Gateway data changes and obtains the mobile metadata it requires to replicate changes.
 Any link:glossary.adoc#non-sync-gateway[non-Sync Gateway change] is eligible for import.
 
-NOTE: By default Sync Gateway distributes import processing across all nodes in the Sync Gateway cluster
+NOTE: By default import processing is switched-off and should only be enabled on one node
```
JFlath: Should we add something version specific, just in case folks don't doublecheck? Or is that messing up the general approach?

```suggestion
NOTE: By default import processing is disabled in version 2.6 and earlier and should only be enabled on one node
```

ibsoln:
@JFlath just needs signoff now

JFlath (approved):

PR #181, opened 2020-02-21
DOC-6304-Import-Docs-Continuous

ibsoln:
https://issues.couchbase.com/browse/DOC-6304

This legacy setting was deprecated in Release-2.1 This ticket completes the task, removing 'continuous' from the example code-snippet of the config that is included in getting started.master.

It will need applying to Releases 2.1-2.7. The 2.1 release notes will also need amending to reflect the deprecation.


ibsoln:
Need to close to add another change (to yaml)

PR #251, opened 2020-09-15
DOC-7355 Apply CBG-873 Feedback to Config and Admin API

ibsoln:
https://issues.couchbase.com/browse/DOC-7355
Note Prometheus changes are covered by https://issues.couchbase.com/browse/DOC-7357

### DIFF
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -8,6 +8,11 @@ properties:
     description: (**Deprecated**) URL of the Sync Gateway Admin Console HTML page.
     type: string
     default: ':4985/_admin/'
+  metricsInterface:
+    description: Port or TCP network address (IP address and the port) which exposes the metrics endpoint.
+    type: string
+    default: none
+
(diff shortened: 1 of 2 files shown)


ibsoln:
Merging to make reviewing easier. @JRascagneres can you still check this please?

JRascagneres:
@ibsoln Sorry, didn't get around to reviewing yesterday. Looks mostly good from the missing fields I gave you. One thing needs changing though and one could maybe be made clearer:
- sgreplicate_websocket_heartbeat_secs - As the name suggests this field is in fact in seconds not in minutes. I think the confusion was down to me saying the default was 5 **mins**. It is 5 minutes but as the config option is provided in seconds it might be better to say default in 300 seconds.
- session_cookie_secure - probably needs to be a bit clearer & changes:
	- So default is true is ssl cert set, otherwise false so this means that if ssl is used secure cert is used (this is currently incorrect, this was my mistake)
	- Otherwise this config can be used to override the default behaviour. Likely use case would be wanting to use insecure cookies but with SSL being used.

PR #354, opened 2020-11-25
DOC-2968 -- Document get/post _changes, timeout and heartbeats min/ma…

ibsoln:
…x ranges

https://issues.couchbase.com/browse/DOC-7769
A sub task of https://issues.couchbase.com/browse/DOC-2968

### DIFF
--- a/modules/ROOT/assets/attachments/sync-gateway-public.yaml
+++ b/modules/ROOT/assets/attachments/sync-gateway-public.yaml
@@ -165,7 +165,7 @@ parameters:
     in: query
     name: heartbeat
     description: |+
-      The `heartbeat` defines the interval (in milliseconds) at which an empty line (CRLF) is written to the response. It helps prevent gateways from deciding the socket is idle and closing it.
+      The `heartbeat` defines the interval (in milliseconds) at which an empty line (CRLF) is written to the response. It helps prevent Sync Gateway from deciding the socket is idle and closing it.
 
       The `heartbeat` value overrides any `timeout` value, to keep the feed alive indefinitely.
 
@@ -177,7 +177,7 @@ parameters:
 
       - Applies ONLY where `feed=longpoll` or `feed=continuous`.
 
-      - Minimum: 25 seconds
+      - Minimum: 25000 (25 seconds)
 
       - Maximum: None -- unless you define one in your configuration file using `MaxHeartbeat`
 
@@ -304,7 +304,7 @@ parameters:
 
       Setting `timeout=0` results in no timeout.
 
-      **Default:** 5 minutes (300 seconds, 300,000 ms)
+      **Default:** 300000 (5 minutes/300 seconds)
 
       **Constraints:**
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/assets/attachments/sync-gateway-public.yaml
```diff
@@ -164,7 +164,23 @@ parameters:
   heartbeat:
     in: query
     name: heartbeat
-    description: Default is 0. Interval in milliseconds at which an empty line (CRLF) is written to the response. This helps prevent gateways from deciding the socket is idle and closing it. Only applicable to longpoll or continuous feeds. Overrides any timeout to keep the feed alive indefinitely. Setting to 0 results in no heartbeat.
+    description: |+
+      The `heartbeat` defines the interval (in milliseconds) at which an empty line (CRLF) is written to the response. It helps prevent gateways from deciding the socket is idle and closing it.
```
djpongh: instead of "gateways" how about "Sync Gateway"

modules/ROOT/assets/attachments/sync-gateway-public.yaml
```diff
@@ -164,7 +164,23 @@ parameters:
...
+
+      Setting `heartbeat=0` results in no heartbeat.
+
+      **Default:** 0, which is no heartbeat
+
+      **Constraints:**
+
+      - Applies ONLY where `feed=longpoll` or `feed=continuous`.
+
+      - Minimum: 25 seconds
```
djpongh: Since the description above says the value is in milliseconds, you should put all values as milliseconds - you can add a more "conventional" units in parens.

EX: 25,000 milliseconds (25 seconds)

modules/ROOT/assets/attachments/sync-gateway-public.yaml
```diff
@@ -358,13 +388,44 @@ parameters:
...
+
+            Setting `heartbeat=0` results in no heartbeat.
+
+            **Default:** 0, which is no heartbeat
+
+            **Constraints:**
+
+            - Applies ONLY where `feed=longpoll` or `feed=continuous`
+
+            - Minimum: 25 seconds
```
djpongh: list as milliseconds

modules/ROOT/assets/attachments/sync-gateway-public.yaml
ibsoln: Makes sense ... applied in e5adab47e55f86e87510f1e15e12642c70806b8b

modules/ROOT/assets/attachments/sync-gateway-public.yaml
ibsoln: Makes sense ... applied in e5adab4

modules/ROOT/assets/attachments/sync-gateway-public.yaml
ibsoln: Makes sense ... applied in e5adab4

djpongh (approved):

PR #361, opened 2020-12-01
DOC-5782 -- Replace swager spec references with ''new' _attachments/s…

ibsoln:
…ync-gateway-public.yaml location

https://issues.couchbase.com/browse/DOC-5782


ibsoln:
Replaced by cherry pick instead

PR #413, opened 2021-02-02
DOC-8078 -- Alternate connection overrides

ibsoln:
https://issues.couchbase.com/browse/DOC-8078
Sync Gateway 2.8.1: DCP feed connections now support connstr ?network

### DIFF
--- a/modules/ROOT/assets/attachments/sg.yaml
+++ b/modules/ROOT/assets/attachments/sg.yaml
@@ -613,7 +613,7 @@ properties:
                   admin_channels:
                     type: array
                     description: |
-                      The list of channels this role is automatically granted access to when Sync Gateway starts.  
+                      The list of channels this role is automatically granted access to when Sync Gateway starts.
                       If "*" is specified then the role is granted access to the star channel which contains all documents.
                     items:
                       type: string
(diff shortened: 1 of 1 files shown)


modules/ROOT/assets/attachments/sg.yaml
```diff
@@ -649,8 +649,25 @@ properties:
 
               As with the SDK, when using the `couchbase://` or `couchbases://` schemes, the port is not required, but if specified should be the external/internal bucket ports (defaults are 11210 or 11207 respectively). Attempting to use the admin ports (8091/18091) will result in a startup error.
 
+              *Alternate Addresses*
+
               On startup, Sync Gateway will try each hostname that is provided until it is able to connect successfully.
 
+              By default, if a remote cluster has an external address set, then when SG connects it will always use the external address.
+              However, it is possible to override this behavior by adding a `network` parameter to the connection string.
```
bbrks: Unfortunately it's not this simple for the `auto` case (SG's default) - We apply a heuristic to determine whether to choose between external or default (internal).

This is based on the hostnames supplied in the connection string: SG will use external networking if the hostname doesn't match Couchbase Server's internal node addresses, and there's an external address defined. Failing that, default (internal) networking is used.

Adding to the confusion, the SDK/SG default of `auto` is not the same as `network=default`!  But that's outside of our control :( 
So need to be very selective when using either of those words to avoid misinterpretation.

bbrks (dismissed):
Left a comment related to how the 'auto' behaviour works.

modules/ROOT/assets/attachments/sg.yaml
ibsoln: Ok. Cheers. Will re-work this tomorrow. Thanks @bbrks

ibsoln: Does this work better @bbrks ?

`By default, if a remote cluster has an external address set, then when SG connects it will apply a heuristic to determine whether to choose between external or default (internal).

The choice is based on the host names supplied in the connection string. 
 - SG uses external networking only when none of the supplied host names match any of Couchbase Server's internal node addresses, and an external address is defined. 
- In all other cases Sync Gateway uses the default (internal) networking.
              
However, it is possible to override this behavior by adding a `network` parameter to the connection string.

The `network` parameter can be --
- auto -- this is the default value if no parameter is provided. In this case the heuristic described above is applied to determine the address used; so effectively there is no override.
- external -- to always force use of the external address
- default -- to always force use of the internal address`

bbrks: Yes, this sounds good. Better than I would have been able to phrase it! 😄

ibsoln: Cheers. It's good to go then.

bbrks (approved):

ibsoln:
Approved change merged into feature/release281 - so closing this PR

PR #493, opened 2021-04-30
DOC-8391 - Server Compatibiliy - Transactions

ibsoln:
https://issues.couchbase.com/browse/DOC-8391

### DIFF
--- a/modules/ROOT/pages/server-compatibility-transactions.adoc
+++ b/modules/ROOT/pages/server-compatibility-transactions.adoc
@@ -23,7 +23,7 @@ include::partial$block-abstract.adoc[]
 
 Couchbase Server provides the backing data store for Sync Gateway.
 
-// TIP: See: {compatibility--xref} for version compatibility information.
+TIP: See: {compatibility--xref} for version compatibility information.
 
 Both Couchbase Server and Couchbase mobile (Sync Gateway and Couchbase Lite) support transactions.
 Here we look at how they relate.
--- a/modules/ROOT/pages/server-compatibility-transactions.adoc
+++ b/modules/ROOT/pages/server-compatibility-transactions.adoc
@@ -3,4 +3,44 @@
 :idprefix:
 :idseparator: -
 
+
+include::partial$_std-hdr-sgw.adoc[]
+
+
+// BEGIN -- Page Attributes
+:url-api-method-database-in-batch: http://docs.couchbase.com/mobile/{version-full}/couchbase-lite-swift/Classes/Database.html#/s:18CouchbaseLiteSwift8DatabaseC7inBatch5usingyyyKXE_tKF[Database.inBatch()]
+// END -- Page Attributes
+
+
+// BEGIN -- Page Heading
+:param-topic-group: compatibility
+:param-related: {compatibility--xref}
+include::partial$block-abstract.adoc[]
+// END -- Page Heading
+
+
+== Introduction
+
+Couchbase Server provides the backing data store for Sync Gateway and it is compatible -- see: {compatibility--xref}.
+However, there are considerations you needs to be aware of when using some Couchbase Server features with Sync Gateway.
+One of those instances concerns the use of Couchbase Server Transactions.
+
+
+== Local Transactions
+
+Within the Couchbase Mobile cluster all transactions are local, either to the Couchbase Server or to the Couchbase Lite database:
+
+* Couchbase Server supports server side transactions -- see: {server-transactions--xref}. +
+Changes made on the server within a transaction block are guaranteed to be imported by the Sync Gateway only after all the changes in a transaction are committed.
+
+* Similarly, Couchbase Lite supports transactions that are local to the Couchbase Lite database -- see: {url-api-method-database-in-batch}. +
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/server-compatibility-transactions.adoc
```diff
@@ -3,4 +3,44 @@
...
+// BEGIN -- Page Heading
+:param-topic-group: compatibility
+:param-related: {compatibility--xref}
+include::partial$block-abstract.adoc[]
+// END -- Page Heading
+
+
+== Introduction
+
+Couchbase Server provides the backing data store for Sync Gateway and it is compatible -- see: {compatibility--xref}.
```
adamcfraser: "it is compatible" is an odd way to describe this - as you say, Couchbase Server is a dependency for Sync Gateway, so stating it's compatible with server (as a whole) doesn't really make sense.

modules/ROOT/pages/server-compatibility-transactions.adoc
```diff
@@ -3,4 +3,44 @@
...
+However, there are considerations you needs to be aware of when using some Couchbase Server features with Sync Gateway.
+One of those instances concerns the use of Couchbase Server Transactions.
```
adamcfraser: This leaves me wondering what the other instances are.  In general I feel like this introduction could be rewritten along the lines of 'Couchbase mobile and server both support transactions, here are details on how they relate'

modules/ROOT/pages/server-compatibility-transactions.adoc
```diff
@@ -3,4 +3,44 @@
...
+
+
+== Local Transactions
+
+Within the Couchbase Mobile cluster all transactions are local, either to the Couchbase Server or to the Couchbase Lite database:
```
adamcfraser: I don't think we really talk about a Couchbase Server+Mobile deployment as a Couchbase Mobile Cluster - at least that's not my usual understanding of a 'cluster'.  Maybe rephrase as 'Couchbase Mobile deployment' or 'Couchbase Mobile ecosystem'?

modules/ROOT/pages/server-compatibility-transactions.adoc
```diff
@@ -3,4 +3,44 @@
...
+
+* Couchbase Server supports server side transactions -- see: {server-transactions--xref}. +
+Changes made on the server within a transaction block are guaranteed to be imported by the Sync Gateway only after all the changes in a transaction are committed.
+
+* Similarly, Couchbase Lite supports transactions that are local to the Couchbase Lite database -- see: {url-api-method-database-in-batch}. +
+Only committed transactions will ever reach Sync Gateway.
+
+Sync Gateway is not transaction-aware.
```
adamcfraser: You've already stated that transactions won't be imported by Sync Gateway until they are committed, so I don't think it's correct to say that Sync Gateway is not transaction aware.  I think the more important clarification is that mobile replication isn't transaction-aware, or more specifically doesn't replicate transactions atomically.  This isn't just a Sync Gateway consideration - Couchbase Lite doesn't replicate documents transactionally either (to SG, or P2P).

modules/ROOT/pages/server-compatibility-transactions.adoc
```diff
@@ -3,4 +3,44 @@
...
+There is no guarantee that a series of updates made within a transaction block, on either the Couchbase Server or Couchbase Lite client side, will be ACID compliant when they sync to the other end.
+
+If there is a use-case to build end-to-end transactions between Couchbase Server and Couchbase Lite clients, you would implement it within the application layer.
+Perhaps by having a separate _control document_ of transaction log entries, which must be synced from the clients that capture the list of related changes.
```
adamcfraser: I don't know about proposing this hypothetical approach - wouldn't it be better to limit the docs to describing the functionality we do support?

adamcfraser (changes requested):

modules/ROOT/pages/server-compatibility-transactions.adoc
ibsoln: Agreed @adamcfraser  I will amend in line with comments before publishing.

modules/ROOT/pages/server-compatibility-transactions.adoc
ibsoln: @adamcfraser  I will amend in line with comments before publishing.

modules/ROOT/pages/server-compatibility-transactions.adoc
ibsoln: Ok. I'll amend ito ecosystem before publishing.

modules/ROOT/pages/server-compatibility-transactions.adoc
ibsoln: @adamcfraser -- changed in line in published version..

modules/ROOT/pages/server-compatibility-transactions.adoc
ibsoln: removed

ibsoln:
Feedback applied prior to publishing

PR #540, opened 2021-08-16
QF210816-8203 -- Diff file for 8203 changes to autopurge page

ibsoln:
(cherry picked from commit 24a8368de13732fbbe047aea1110cfd9742a4054)

### DIFF
--- a/mydiff.diff
+++ b/mydiff.diff
@@ -1,0 +1,249 @@
+diff --git a/modules/ROOT/pages/auto-purge-channel-access-revocation.adoc b/modules/ROOT/pages/auto-purge-channel-access-revocation.adoc
+new file mode 100644
+index 00000000..ee2af1fb
+--- /dev/null
++++ b/modules/ROOT/pages/auto-purge-channel-access-revocation.adoc
+@@ -0,0 +1,243 @@
++= Auto-Purge on Channel Access Revocation
++:page-aliases: anchor-auto-purge-channel-access-revocation
++:page-edition: 3.0
++:description: pass:q[Auto-purge behavior on loss of access to document channels]
++:keywords: access control, document routing, sync, auto-purge
++
++include::partial$_std-hdr-sgw.adoc[]
++
++
++// BEGIN -- Page Attributes
++:SGW: pass:q[_Sync Gateway_]
++// END -- Page Attributes
++
++
++//  BEGIN -- Page Heading
++:param-topic-group: access-control
++:param-abstract!:
++:param-related!:
++include::partial$block-abstract.adoc[]
++//  END -- Page Heading
++
++== Overview
++
++// If the document was previously routed to a channel, but the current call to the sync function (for an updated revision) doesn't route it to that channel, the document is removed from the channel.
++
++// This may cause users to lose access to that document -- see <<lbl-access-revocation>> for the revocation behavior.
++
++
++// :param-page: replication
++// :param-bookmark: replication-events
++// include::partial$blocklinks-cbl.adoc[]
++
++
++// [#lbl-access-revocation]
++// == Access Revocation
++
++
++Users may lose access to documents for many reasons, including:
++
++* The User loses direct access to channel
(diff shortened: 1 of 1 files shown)


mydiff.diff
```diff
@@ -0,0 +1,249 @@
...
++|Passive Sync Gateway (Remote)
++|Expected Sync behavior
++
++|Pull
++|User revoked access to channel
++|Previously synced documents are auto purged on local
++
++|Push
++|User revoked access to channel
++|Previously synced documents are auto purged on local
```
JRascagneres, line 177: Push replications do not result in any revocations or purges occurring.

mydiff.diff
```diff
@@ -0,0 +1,249 @@
...
++
++
++|PushAndPull
++|User revoked access to channel +
++Sync Function includes `requireAccess(“channel”)`
```
JRascagneres, line 182: requireAccess only limits the ability for clients to push to a channel they don't own. Has no affect on the new revocation work.

mydiff.diff
```diff
@@ -0,0 +1,249 @@
...
++|When access is revoked on remote, the previously synced documents for User2 are auto-purged on local.
++
++Local changes continue to  be pushed to remote but rejected by remote
```
JRascagneres, line 185: I suppose this is related to `requireAccess`

mydiff.diff
```diff
@@ -0,0 +1,249 @@
...
++
++|===
++====
++
++
++=== Access Regained
++
++If a user subsequently regains access to a lost channel then any previously auto-purged documents still assigned to any of their channels are automatically pulled down by the active Sync Gateway  -- see: <<lbl-isgr-regain-behaviors>>.
++
++If you want to control whether to sync previous auto purged versions of documents (rather than pull down purged documents) then you must also move the documents out of all of the users' channels so they are not synced down again.
```
JRascagneres, line 195: Same as on CBL side, this is confusing? Not sure what this is supposed to mean.

mydiff.diff
```diff
@@ -0,0 +1,249 @@
...
++// .2+^.^h
++|Direction
++|Passive Sync Gateway (Remote)
++|Expected Sync behavior
++
++|Pull
++|User REASSIGNED access to channel
++
++Sync Function includes
++`requireAccess( reassignedChannel)`
```
JRascagneres, line 217: Not required

JRascagneres (commented):
Generally fine but not sure about the use of `requireAccess` in both the SGW and CBL docs. It can be used in this context but it is unrelated to the revocation work, it has the same behaviour as it had prior to this work. Not sure if it just adds confusion.

Additionally there is a point I'd like to get included in the docs:

It should be noted that when this option is turned on it only applies to revocations that occur after the point the replication is up to. It should be stated therefore that if this option is turned on from off on an existing replication, revocations that occurred before that point will not be retro-actively purged. To have this done a reset must be carried out. On CBL side a reset checkpoint must be carried out.

mydiff.diff
ibsoln: Fixed

mydiff.diff
ibsoln: Removed

mydiff.diff
ibsoln: Removed sync functoin

mydiff.diff
ibsoln: Removed sync function

mydiff.diff
ibsoln: > Generally fine but not sure about the use of `requireAccess` in both the SGW and CBL docs.

It was included because it is repeatedly mentioned in PRD. I  have removed mention of the sync function in this context now.

ibsoln:
@JRascagneres  > Generally fine but not sure about the use of `requireAccess` in both the SGW and CBL docs.

It was included because it is repeatedly mentioned in PRD, but I  have removed mention of the sync function in this context now.

PR #542, opened 2021-08-16
DOC-8211-C3 -- Re-tweak API accss page

ibsoln:
https://issues.couchbase.com/browse/DOC-8211

DOC-8211-C3 -- Re-tweak API accss page
https://issues.couchbase.com/browse/DOC-8211

DOC-8211-C3 -- Re-tweak API accss page
https://issues.couchbase.com/browse/DOC-8211

### DIFF
--- a/modules/ROOT/pages/rest-api-access.adoc
+++ b/modules/ROOT/pages/rest-api-access.adoc
@@ -12,28 +12,42 @@ include::partial$_std-hdr-sgw.adoc[]
 include::partial$block-abstract.adoc[]
 
 :xref--pfx: {sgw--xref}:
+:tls-encryption--xref: {sgw--xref}{secure-sgw-access--page}[enforcing TLS encryption]
 
 == Overview
 
-In addition to the {rest-api--xref}, Sync Gateway provides secure access to two REST API  targeted at specific user populations and functional areas, namely the:
+Sync Gateway provides secure access to its REST APIs, namely the:
 
-* {rest-api-admin--xref}
-* {rest-api-metrics--xref}
+* {rest-api--xref}
+* {rest-api-admin--xref} -- for the administration and configuration of {sgw}
+* {rest-api-metrics--xref} -- for the monitoring of {sgw} performance metrics
 
 Each REST API is accessed through a different, <<lbl-port-cfg, user-specifiable>>, TCP port.
-This makes it easy to control their physical exposure, perhaps keeping the Admin REST API secure behind your firewall.
+This makes it easy to control their physical exposure, perhaps to keep the Admin REST API secure behind your firewall.
 
-{sgw-s} 3.0 brings additional layers of security to the Metrics and Admin interfaces by requiring Couchbase Server Role-Based Access Control (RBAC) authentication for all <<lbl-secure-users,secure API users>>, and by {sgw--xref}{security--page}[enforcing TLS encryption].
+{sgw-s} 3.0 brings additional layers of security by {tls-encryption--xref} for all API traffic and by requiring Couchbase Server Role-Based Access Control (RBAC) authentication for all <<lbl-secure-users,Admin and Metrics API users>>.
 
 RBAC user authentication enables secure remote administration of {sgw} clusters.
 This is critical in cloud native deployments.
 It provides secure and fine-grained access control.
 
 
-[#lbl-secure-users]
-== Secure API Users
+== Basic Authentication
 
-In order to submit Admin or Metrics REST API requests you must create Couchbase Server users for that purpose, and to then provide a valid set of Couchbase Server credentials for the pre-configured Couchbase Server user on each API request.
+In 3.0, TLS is enabled by default.
+It will use the use credentials provided with the request for authentication.
+That will be the {sgw} user on the Public REST API and the appropriated Couchbase Server RBAC-user credentials for Metrics and Admin APIs.
+
+The only exception to this is the use of the Public REST API's _GUEST_ mode; which is for use in testing and-or development environments only.
+
+For more on TLS, see: {tls-encryption--xref}.
+
+
+[#lbl-rest-users]
+== RBAC Users
+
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/rest-api-access.adoc
```diff
@@ -12,28 +12,42 @@ include::partial$_std-hdr-sgw.adoc[]
...
+* {rest-api--xref}
+* {rest-api-admin--xref} -- for the administration and configuration of {sgw}
+* {rest-api-metrics--xref} -- for the monitoring of {sgw} performance metrics
 
 Each REST API is accessed through a different, <<lbl-port-cfg, user-specifiable>>, TCP port.
-This makes it easy to control their physical exposure, perhaps keeping the Admin REST API secure behind your firewall.
+This makes it easy to control their physical exposure, perhaps to keep the Admin REST API secure behind your firewall.
 
-{sgw-s} 3.0 brings additional layers of security to the Metrics and Admin interfaces by requiring Couchbase Server Role-Based Access Control (RBAC) authentication for all <<lbl-secure-users,secure API users>>, and by {sgw--xref}{security--page}[enforcing TLS encryption].
+{sgw-s} 3.0 brings additional layers of security by {tls-encryption--xref} for all API traffic and by requiring Couchbase Server Role-Based Access Control (RBAC) authentication for all <<lbl-secure-users,Admin and Metrics API users>>.
```
JRascagneres, line 28: On the staging site {sgw-s} is just that in plain text. Not sure if thats just because its in staging or if there is an issue with it.

JRascagneres: Additionally I know you specify later but might be worth noting that TLS isn't really "enforced". Its enabled by default but a customer can turn it off. Same with the Admin Auth, its not 'required' its optional on the customers part.

modules/ROOT/pages/rest-api-access.adoc
```diff
@@ -12,28 +12,42 @@ include::partial$_std-hdr-sgw.adoc[]
...
 This is critical in cloud native deployments.
 It provides secure and fine-grained access control.
 
 
-[#lbl-secure-users]
-== Secure API Users
+== Basic Authentication
 
-In order to submit Admin or Metrics REST API requests you must create Couchbase Server users for that purpose, and to then provide a valid set of Couchbase Server credentials for the pre-configured Couchbase Server user on each API request.
+In 3.0, TLS is enabled by default.
```
JRascagneres, line 37: This sounds like we're confusing TLS and admin auth. 
TLS is on the same topic of security but separate from the admin auth work. You can have admin auth without TLS and TLS without admin auth.

modules/ROOT/pages/rest-api-access.adoc
```diff
@@ -12,28 +12,42 @@ include::partial$_std-hdr-sgw.adoc[]
...
+It will use the use credentials provided with the request for authentication.
+That will be the {sgw} user on the Public REST API and the appropriated Couchbase Server RBAC-user credentials for Metrics and Admin APIs.
```
JRascagneres, line 39: appropriated --> appropriate ?

modules/ROOT/pages/rest-api-access.adoc
```diff
@@ -54,7 +68,7 @@ For more on creating Couchbase Server users see the Couchbase Server content her
```
JRascagneres, line 68: This is a few lines down as I couldn't access the lines further up in this PR but:

```
Authenticated users will have full access to admin API functionality, application data and configuration settings, depending upon the selected role.
```

Maybe drop 'full' from 'full access' kind of sounds like its contradicting itself when you say 'have full access' but then 'depending upon role'.

JRascagneres:
```
{enterprise} users can exert finer-grained control over functionality by using  <<lbl-rbac-roles>>
```

on the site this reads as
```users can exert finer-grained control over functionality by using RBAC Roles```

this doesn't quite read right as even non-enterprise users can use RBAC roles as you say above, they just have less to choose from. Maybe:

```
users can exert finer-grained control over functionality by using the larger selection of enterprise RBAC Roles
```
I'll let you come up with wording but imo that makes more sense

File: modules/ROOT/pages/rest-api-access.adoc [hunk 5]
```diff
@@ -54,7 +68,7 @@ For more on creating Couchbase Server users see the Couchbase Server content her
 
 
 [#lbl-rbac-roles]
-== Authentication Roles
+== RBAC Roles
 
 include::partial$block-caveats.adoc[tags=ee-only-content]
```
JRascagneres, line 73: Again this is more applicable to stuff below
image

image

Got some weird overlap.

JRascagneres (commented):
Did the best I could to apply comments to this. It's a bit difficult with the changes split across multiple PRs so attempted to put them on here.

---

Couple larger points:
- Port configuration table doesn't seem to let me switch tabs.

- Config option `enable_advanced_auth_dp` should be put somewhere. This is for the finer-grained EE permissions - By default this will be off. Its a bool taking true / false. **The finer grained roles are developer preview in 3.0**

For the Admin REST API page:

```
To allow users to access the Admin API up you need to create a Couchbase Server-based RBAC-user for them — see: REST API Access
```

Unless admin auth is disabled.

For the Metrics REST API page:
 - I don't see any changes here. Is that intentional?
 
---

I think we should make it clear that admin auth is optional too. We do mention you can disable in some places but everything is written as 'you have to'. Running without this is valid.

The initial block also seems to confuse TLS and admin auth. We should make sure these are separated

modules/ROOT/pages/rest-api-access.adoc
ibsoln: > On the staging site {sgw-s} is just that in plain text. Not sure if thats just because its in staging or if there is an issue with it.

There is a staging site issue causing these shared variables not to resolve.

PR #693, opened 2022-05-23
DOC-10076 Remove DP label from SGW RBAC roles

RayOffiah:
DP tag removed

### DIFF
--- a/modules/ROOT/pages/rest-api-access.adoc
+++ b/modules/ROOT/pages/rest-api-access.adoc
@@ -84,7 +84,7 @@ Note that the only role available for community-edition users is the *Full Admin
 
 h| Role
 h| Capability
-h| 7.02
+h| 7.0.2{plus}
 h| 6.1 - 7.0
 h| 5.5 - 6.0
 h| 5.1
--- a/modules/ROOT/pages/rest-api-access.adoc
+++ b/modules/ROOT/pages/rest-api-access.adoc
@@ -79,12 +79,13 @@ Note that the only role available for community-edition users is the *Full Admin
 
 
 .{sgw-s} role availability by release
-[#tbl-ee-svr-sgw-roles,cols="2,2,^1,^1,^1,^1",options="header"]
+[#tbl-ee-svr-sgw-roles,cols="2,2,^1,^1,^1,^1,^1",options="header"]
 |===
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/rest-api-access.adoc
```diff
@@ -84,7 +84,7 @@ Note that the only role available for community-edition users is the *Full Admin
 
 h| Role
 h| Capability
-h| 7.02 DP
+h| 7.02
```
markspolakovs: Should this be 7.0.2 rather than 7.02? Also perhaps worth saying something like "7.0.2 and up" (just to head off the inevitable "what about 7.1" questions).

RayOffiah: Yup, good point.

RayOffiah:
@markspolakovs 

How's this?

markspolakovs:
@RayOffiah Sorry, chatted through this with the team and it seems like I had it a bit backwards - so my understanding is that the roles are available in 7.0.2-7.0.x behind Developer Preview, and 7.1.0 by default (no DP). Apologies!

markspolakovs:
Perhaps also worth linking to https://docs.couchbase.com/server/current/developer-preview/preview-mode.html, just to explain what DP is for people who only see the SGW docs (not the Server docs)?

RayOffiah:
@markspolakovs 

New changes.

modules/ROOT/pages/rest-api-access.adoc
```diff
@@ -79,12 +79,13 @@ Note that the only role available for community-edition users is the *Full Admin
...
 
 .{sgw-s} role availability by release
-[#tbl-ee-svr-sgw-roles,cols="2,2,^1,^1,^1,^1",options="header"]
+[#tbl-ee-svr-sgw-roles,cols="2,2,^1,^1,^1,^1,^1",options="header"]
 |===
 
 h| Role
 h| Capability
-h| 7.0.2{plus}
+h| 7.10
```
markspolakovs:
```suggestion
h| 7.1.0
```

markspolakovs (approved):

osfameron (approved):
👍

PR #697, opened 2022-06-06
Update 3.0.3 release notes

ThreeJackSparrow:
Added CBG-2055 and a note that upgrade to 3.0.3 is strongly recommended.


osfameron (approved):
👍  I understand (ish) that there is no 3.0.2/3.0.1 release

PR #700, opened 2022-06-13
3.0.3 replaces .0

ThreeJackSparrow:
Add notes to the release notes and update the compatibility tables to show that v3.0.0 is replaced by v3.0.3.


osfameron (approved):
👍

PR #702, opened 2022-06-14
DOC-10130 Errors on Sync Gateway Supported Environments Page

RayOffiah:
https://issues.couchbase.com/browse/DOC-10130

Fixed

### DIFF
--- a/modules/ROOT/pages/_partials/sgw-supported-os.adoc
+++ b/modules/ROOT/pages/_partials/sgw-supported-os.adoc
@@ -78,6 +78,10 @@ a|Ubuntu
 
 | Windows
 | Desktop 10
+
+| Apple M1 ARM64
+| macOS 11
+
 |===
 // end::sup-os-dev-test[]
 // end::supported[]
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/_partials/sgw-supported-os.adoc
```diff
@@ -26,8 +26,7 @@
 a| 2019 +
 2016 (64-bit) -- deprecated at {version-full}
 
-| Apple M1 ARM64
-| macOS 11
+
```
djpongh, line 29: this should be moved down to table 3 (need to explicitly call out Arm64 )

djpongh (changes requested):

modules/ROOT/pages/_partials/sgw-supported-os.adoc
RayOffiah: Fixed

RayOffiah:
@djpongh 

Could you take another look at the changes? 

The page doesn't usually mention the hardware under the operating system, so it does stand out a bit.

Thanks.

djpongh:
looks good - Apple is a special case because their old systems are x86 based and the new systems are ARM64 based

djpongh (approved):

PR #704, opened 2022-07-11
DOC-10227-Couchbase (Server) Requirements: Query and Index services as prerequisites for Sync Gateway

RayOffiah:
https://issues.couchbase.com/browse/DOC-10227

Added section for the cluster requirements.
Tidied up some of the numbering.


tonyjhillman (approved):

modules/ROOT/pages/get-started-prepare.adoc
```diff
@@ -40,10 +40,15 @@ Here's what you need in order to install Sync Gateway:
 
 Once you have all that covered ... go {get-started-install--xref} {sgw-t}.
 
+== Cluster Cluster Requirements
```
YoshiyukiKono: This must be a typo. "**Cluster** Cluster Requirements" should be "Couchbase Cluster Requirements".

Please also see below comments. The place of this chapter (and the level of chapter) seems misleading to readers.

modules/ROOT/pages/get-started-prepare.adoc
```diff
@@ -40,10 +40,15 @@ Here's what you need in order to install Sync Gateway:
...
+Couchbase clusters intended for use with Sync Gateway must be configured with the following services:
 
+* The `Data` Service.
+* The `Index` Service.
+* The `Query` Service.
 
 [#lbl-req-minim, reftext=Minimum Requirements]
-== Couchbase Requirements
+== Couchbase Server  Requirements
```
YoshiyukiKono: There are two blank characters between "Server" and "Requirements".

modules/ROOT/pages/get-started-prepare.adoc
```diff
@@ -40,10 +40,15 @@ Here's what you need in order to install Sync Gateway:
 
 Once you have all that covered ... go {get-started-install--xref} {sgw-t}.
 
+== Cluster Cluster Requirements
+Couchbase clusters intended for use with Sync Gateway must be configured with the following services:
 
+* The `Data` Service.
+* The `Index` Service.
+* The `Query` Service.
```
YoshiyukiKono: For readers, the entire procedure should be in the following order.
1. Recognize the dependency to another software (Couchbase Server).
2. Download the software (if needed). **You can get Couchbase Server from our [Downloads](https://www.couchbase.com/downloads/?family=mobile) page**
3. Setup a cluster with understanding **Couchbase Cluster Requirements**, i.e. Data, Index, Query.
4. **You will then need to configure Couchbase Server by adding a Bucket and an RBAC User for Sync Gateway**

So, this part (3rd step) should be between 2nd and 4th. It is more natural that "Cluster Requirements" is a sub chapter of "Couchbase Server Requirements" rather than precedent chapter before it. However, there is no other sub chapters there. I may propose to move this part to appropriate place without using its chapter name (...Cluster Requirements). Just an idea.

YoshiyukiKono (changes requested):
Please confirm my comments.

modules/ROOT/pages/get-started-prepare.adoc
RayOffiah: Yup, definitely a typo.

modules/ROOT/pages/get-started-prepare.adoc
RayOffiah: Cleaned up, though this doesn't cause a problem since Asciidoc will render it as a single space when the pages are built.

YoshiyukiKono (approved):
I confirmed that changes have been made to resolve the issues I pointed out.

PR #705, opened 2022-07-28
[DOC-10271] Sync Gateway 3.0.0+ is incompatible with CB Server <6.5.0

RayOffiah:
Changed table to show that CB < 6.6 is incompatible with SGW 3.0x


markspolakovs (commented):
I'm not entirely sure how these table columns line up, but this would suggest that 6.0-6.5 are not compatible with 3.0? 6.5 is, it's only 5.5 and 6.0 that are not.

markspolakovs (approved):

PR #715, opened 2022-11-10
DOC-10537 Broken log troubleshooting links in "Connect to Sync Gateway" verify step

RayOffiah:
https://issues.couchbase.com/browse/DOC-10537


osfameron (approved):
Weird. Wonder why this was abstracted into a variable, but looks good 👍

PR #725, opened 2023-03-13
DOC-10404: Updates Import Filter Configuration URL

franrob-projects:
Updates Import Filter Configuration URL from `/{db}/` to `/{keyspace}/`

..and changes xref type to something more commonly understandable throughout the team

### DIFF
--- a/modules/ROOT/pages/import-filter.adoc
+++ b/modules/ROOT/pages/import-filter.adoc
@@ -53,7 +53,7 @@ Without a filter (the default), Sync Gateway imports all documents and so we rec
 
 == Function Provision
 
-Use the {rest-api-admin--xref--database-configuration} Admin Rest API endpoint xref:clusters:modules/Admin REST API[/{db}/_config/import_filter] to provision an import filter for a database using the `application/javascript` mime type.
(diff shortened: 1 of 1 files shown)


dinaabennett (approved):
:-)

modules/ROOT/pages/import-filter.adoc
```diff
@@ -53,7 +53,7 @@ Without a filter (the default), Sync Gateway imports all documents and so we rec
 
 == Function Provision
 
-Use the {rest-api-admin--xref--database-configuration} Admin Rest API endpoint {rest-api-admin--xref--import-filter} to provision an import filter for a database using the `application/javascript` mime type.
+Use the {rest-api-admin--xref--database-configuration} Admin Rest API endpoint xref:clusters:modules/Admin REST API[/{keyspace}/_config/import_filter] to provision an import filter for a database using the `application/javascript` mime type.
```
osfameron, line 56: Does this work and give the same result as previously? e.g. point to the same URL fragment etc.?

"Admin REST API" looks odd in an xref macro.
It does work in `<< >>` macros so I guess it will do?
Though those aren't the best option - see https://docs.asciidoctor.org/asciidoc/latest/macros/xref/#natural-cross-reference

PR #729, opened 2023-04-06
Doc 10993  Review mobile helium docs

franrob-projects:
Reviews and corrects feedback on the following:

DOC-10993: Removes incorrect function from Sync Function page

DOC-10993: Removes incorrect function from Auto-Purge page
 
DOC-10993: Removes incorrect function from Acess Control page
 
DOC-10993: Removes incorrect function from Channels page
 
DOC-10993: Removes incorrect function from Sync-Function-API page
 
DOC-10993: Corrects typo
 
DOC-10993: Adds /{keyspace}/ in the place of /{db}/

DOC-10993: Fixes link syntax - must check in preview also


osfameron (approved):
👍

PR #748, opened 2023-06-06
Doc 11143:  Fixes broken asciisoc syntax, remove s&c content 

franrob-projects:
- Fixes broken asciisoc syntax
- Remove s&c content

### DIFF
--- a/modules/ROOT/pages/data-modeling.adoc
+++ b/modules/ROOT/pages/data-modeling.adoc
@@ -29,20 +29,20 @@ They do not necessarily align with constraints on the local storage and use of s
 == Property Naming
 
 You can use an underscore prefix (`+_+`, ASCII `&#095`) for property naming, but your name cannot match any of the _Document_ system properties reserved by Sync Gateway:
-+
-* _sync
-* _id
-* _rev
-* _deleted
-* _attachments
-* _revisions
-* _exp
-* _purged
-* _removed
+
+* `_sync`
+* `_id`
+* `_rev`
+* `_deleted`
+* `_attachments`
+* `_revisions`
+* `_exp`
+* `_purged`
+* `_removed`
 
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/data-modeling.adoc
```diff
@@ -73,19 +73,12 @@ You should change any top-level user properties that have a key with a leading u
 * Renaming them to remove the underscore, or,
 * Wrapping them inside another object with a key that doesn't have a leading underscore.
 
-
-
-
-
-
```
osfameron, line 80: Deleting multiple whitespace lines is fine, but just to note, sometimes people find them useful before a `==` to help break up the structure.
I think *2* newlines is sufficient in that case though.

(Skim-reading an asciidoc is hard enough as it is.)

Anyway, just a thought to put that into context.

modules/ROOT/pages/data-modeling.adoc
```diff
@@ -127,25 +120,6 @@ Its main purpose is to enable the replicator to determine which data to sync and
 
 Each document change is assigned a unique revision ID. The IDs of past revisions are available. The content of past revisions may be available if the revision was created locally and the database has not yet been compacted.
 
-[[Scopes-and-Collections]]
-== Scopes and Collections
```
osfameron, line 131: I'm not quite sure why you removed this, but I think you mentioned PM was in agreement?
If so, then 👍

osfameron (approved):
👍

modules/ROOT/pages/data-modeling.adoc
franrob-projects: Yeah I got some feedback that this was not in the right place. So it's all been shifted to 'whats new' 'Confg' and 'Sync'

PR #763, opened 2023-09-08
Doc 11333 sgcollect

osfameron:
https://issues.couchbase.com/browse/DOC-11333

Added username and password, and also log-redaction-level, as this was supplied in a usage example from @mohammed-madi

### DIFF
--- a/modules/ROOT/pages/sgcollect-info.adoc
+++ b/modules/ROOT/pages/sgcollect-info.adoc
@@ -70,6 +70,9 @@ Collect Sync Gateway diagnostics and upload them to Couchbase Support:
 ./sgcollect_info \
   --sync-gateway-config=/path/to/config.json \
   --sync-gateway-executable=/usr/bin/sync_gateway \
+  --sync-gateway-username=Admin \
+  --sync-gateway-password=password \
+  --log-redaction-level=partial \
   --upload-host=uploads.couchbase.com \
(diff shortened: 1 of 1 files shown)


mohammed-madi (changes requested):
The issue we have is that it is not clear on https://docs.couchbase.com/sync-gateway/current/sgcollect-info.html that the `--sync-gateway-password `and `--sync-gateway-username` flags are needed when admin_interface_authentication is set to true in the config, and running sgcollect without them will lead to pprof files not being collected as the calls to sgw metrics endpoints would require admin auth (when admin_interface_authentication is set to true) and sgcollect can only get it from those flags when ran from the command line. can we get a warning window or something similar with that explained please?
https://issues.couchbase.com/browse/DOC-10860 for more info

osfameron:
Thanks @mohammed-madi that looks clear, unfortunately I've timed out and am moving house next week.
@RichardSmedley (cc'd) may be able to assist in my absence.

Note that if you're able to contribute a commit to this PR, this would be incredibly helpful.

mohammed-madi (approved):
It would be nice to have the warning but the current change is sufficient for the urgency of this issue.

RichardSmedley:
Thanks  @mohammed-madi - I'll merge and build.
Will try and free up some time next week to take a look at the warning.  :+1:

PR #773, opened 2024-01-05
DOC-11503: fixed bucket description

ElliotFrancisHunter:
https://issues.couchbase.com/browse/DOC-11503

Fixed issue below:
Bucket Key description says
Sync Gateway interactions with the bucket (trace level only).

As of 3.1 this has been moved to debug level, please remove " (trace level only)."


osfameron (approved):
👍

PR #774, opened 2024-01-08
DOC-11517: correct port for create new document

ElliotFrancisHunter:
https://issues.couchbase.com/browse/DOC-11517

This is a fix for the above ticket. Description as follows: 
"Create a New Document"

Uses the wrong port. The Admin API is not the API used for Sync Server users call so should target http://localhost:4984/ not http://localhost:4985/


osfameron (approved):
👍

PR #775, opened 2024-01-08
[Backport 3.0] DOC-11517: Corrected port to 4984

ElliotFrancisHunter:
https://issues.couchbase.com/browse/DOC-11517

This is a fix for the above ticket. Description as follows:
"Create a New Document"

Uses the wrong port. The Admin API is not the API used for Sync Server users call so should target http://localhost:4984/ not http://localhost:4985/

### DIFF
--- a/modules/ROOT/pages/get-started-verify-install.adoc
+++ b/modules/ROOT/pages/get-started-verify-install.adoc
@@ -272,7 +272,7 @@ Within a terminal use CURL to issue the following POST request, which adds a new
 ----
 DIGEST=`echo -n sgwuser1:password | base64`
 
-curl  --location --request PUT 'http://localhost:4985/traveldb/hotel_88801' \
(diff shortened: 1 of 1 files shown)


osfameron (changes requested):
This looks good, but aren't the rest of the entries under the CRUD cycle *also* against the public API?
(e.g. the Get a Document using API, Update, Delete ?)

ElliotFrancisHunter:
> This looks good, but aren't the rest of the entries under the CRUD cycle _also_ against the public API? (e.g. the Get a Document using API, Update, Delete ?)

@osfameron you're completely correct. I had a proper read and had a chat with Ben B. about this to check the coverage of this port swap, it started in 3.0 as part of a cleanup of the config files. Will get on this change now!

osfameron (approved):
👍

PR #783, opened 2024-02-29
[BACKPORT] DOC-10808: make the performance impact more prominent.

ElliotFrancisHunter:
This is a fix for the ticket: https://issues.couchbase.com/browse/DOC-10808

Ticket Description: The warning about the negative performance impact of repeated access() calls from the sync function should be made more prominent. This could be moved to the top of the page, be placed in a warning box to emphasise, etc.

Currently the only warning is under the Context subheading.

"But note that, invoking it multiple times to grant the same user access to the same channel, will result in negative performance implications."

Which is quite easy to overlook.

I have moved the quote to the bottom of the description, changed it to warning to increase the prominence and made the warning more specific in relation to the specific performance issues it causes.

### DIFF
--- a/modules/ROOT/pages/_partials/sync-api/sync-function-api-access.adoc
+++ b/modules/ROOT/pages/_partials/sync-api/sync-function-api-access.adoc
@@ -26,7 +26,6 @@ NOTE: As a convenience, the resolved value of either argument may be `null` or `
 == Context
 
 You can invoke this function multiple times from within your Sync Function.
-But note that, invoking it multiple times to grant the same user access to the same channel, will result in negative performance implications.
 
 TIP: Prefix the `username` argument value with `role:` to apply this function to a role rather than a user.
 This grants access to the specified channel(s) for all users assigned that role.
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/_partials/sync-api/sync-function-api-access.adoc
```diff
@@ -58,5 +56,6 @@ access ("snej", null);
 <.> The null arguments mean these are treated as no-ops
 ====
 
+WARNING: Invoking the 'access()' function multiple times to grant the same user access to the same channel, will result in negative performance implications such as abnormally large fetches or request timeouts.
```
sarahlwelton:
```suggestion
WARNING: If you invoke the `access()` function multiple times to grant the same user access to the same channel, you could see negative performance effects, such as large fetches or request timeouts. 
```
If this is a backport, let's make sure we're using the same text.

sarahlwelton (commented):
Still curious about the `pass:q` markup, but please make sure that the text matches in your backport.

modules/ROOT/pages/_partials/sync-api/sync-function-api-access.adoc
ElliotFrancisHunter: Nice, done!

sarahlwelton (approved):
Looks good to me

PR #798, opened 2024-07-11
[BACKPORT] DOC-11833: Update message to legacy.

ElliotFrancisHunter:
This is a backport fix from 3.0. https://github.com/couchbase/docs-sync-gateway/pull/797
Fix to a ticket raised by support: https://issues.couchbase.com/browse/DOC-11833

I've moved the out of date information to a NOTE block mentioning it applies to versions 2.x

Also some tidying up of the file to remove white space.

The following is noted when discussing potential issues with field names with Sync Gateway:

In Mobile-to-Web Data Sync with Field-level Encryption enabled, because the rule conflicts with the default field encryption format

However, this only applies to SDK 2 which made use of an underscored _crypt key prefix. In SDK3 this was changed to encrypted$ to avoid conflicting with Sync Gateway.

While this can be changed to the legacy, saying outright that field level encryption will cause issues with Sync Gateway is no longer accurate.

### DIFF
--- a/modules/ROOT/pages/data-modeling.adoc
+++ b/modules/ROOT/pages/data-modeling.adoc
@@ -7,12 +7,10 @@ ifdef::prerelease[:page-status: {prerelease}]
 
 include::partial$_set_page_context.adoc[]
 
-
 // BEGIN -- Page Attributes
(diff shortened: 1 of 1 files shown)


RichardSmedley (approved):
:+1:

PR #804, opened 2024-08-15
Doc 12438 customer fixes

ElliotFrancisHunter:
These are fixes for the tickets here: https://issues.couchbase.com/browse/DOC-12438

- DOC-12401
- DOC-12409
- DOC-12413
- DOC-12414

Still requires some tweaking of yaml files with @osfameron to ensure the changes to legacy doc have carried through


modules/ROOT/assets/attachments/bundled-metric.yaml
```diff
@@ -183,42 +183,72 @@ paths:
...
                                 type: integer
+                                description: HeapIdle is bytes in idle (unused) spans. Idle spans have no objects in them. These spans could be (and may already have been) returned to the OS, or they can be reused for heap allocations, or they can be reused as stack memory. HeapIdle minus HeapReleased estimates the amount of memory that could be returned to the OS, but is being retained by the runtime so it can grow the heap without requesting more memory from the OS. If this difference is significantly larger than the heap size, it indicates there was a recent transient spike in live heap size.
                               go_memstats_heapinuse:
                                 type: integer
+                                description: HeapInuse is bytes in in-use spans. In-use spans have at least one object in them. These spans an only be used for other objects of roughly the same size. HeapInuse minus HeapAlloc estimates the amount of memory that has been dedicated to particular size classes, but is not currently being used. This is an upper bound on fragmentation, but in general this memory can be reused efficiently.
```
yuvrajkanwar20, line 201: Firstly, I want to suggest the following edit here:
* These spans an only be used -> These spans can only be used

Also, would we able to please provide any of the following details as well on `go_memstats_heapinuse`?
* When will the memory be released?
* What Sync Gateway configuration settings control/affects the behavior of this stat?

modules/ROOT/assets/attachments/configuration-properties-legacy.yaml
```diff
@@ -7,9 +7,12 @@ properties:
     type: boolean
     default: 'true'
     description: |+
-      Use `admin_interface_authentication=false` to turn of the secure remote administration feature introduced in 3.0.
+      This is set to `true` by default and requires https://docs.couchbase.com/sync-gateway/current/rest-api-access.html#lbl-secure-users[Couchbase Server RBAC-users] to authenticate Admin REST API access and send requests to the Admin API.
```
yuvrajkanwar20: Just a small edit suggestion here: 
Admin REST API -> https://docs.couchbase.com/sync-gateway/current/rest-api-admin.html[Admin REST API]

(creating a link to that documentation as well)

ElliotFrancisHunter: Done!

modules/ROOT/assets/attachments/bundled-metric.yaml
ElliotFrancisHunter: This file is generated from API specs created by devs and brought in by us. I think this is out of the scope of this specific change and could be raised as a separate ticket for the sync gateway team. cc @osfameron who handled pulling this down.

yuvrajkanwar20: That sounds good @ElliotFrancisHunter I will bring this up with the team in a different forum and then maybe we can revisit. Thank you for your input.

yuvrajkanwar20 (approved):

osfameron (approved):

PR #806, opened 2024-08-23
[DOC-10197/release/3.1]: Bootstrap config hostnames must be of the Data Service Nodes specifically. Added warning note about using a node running the data service.

RayOffiah:
https://issues.couchbase.com/browse/DOC-10197

### DIFF
--- a/modules/ROOT/assets/attachments/bundled-admin.yaml
+++ b/modules/ROOT/assets/attachments/bundled-admin.yaml
@@ -2267,7 +2267,7 @@ paths:
                         type: string
                         default: 10s
                       server:
-                        description: Couchbase Server connection string/URL.
+                        description: Couchbase Server connection string/URL. (The connection must point to a node running the data service)
                         type: string
                       username:
                         description: Username for authenticating to server.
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/_partials/connection-node-warning.adoc
```diff
@@ -0,0 +1 @@
+IMPORTANT: The connection must point to a node running the data service.
```
ElliotFrancisHunter: Is this an admonition that will see frequent use in the docs? I'm not sure the abstraction is required for two instances

RayOffiah: Probably not, actually; I think I'm only using it in two places. I tend to use includes when the text is going to appear in two or more places.

It's a habit from my programming days, I think. I can certainly bin it though.

RayOffiah: No, hang on; just remembered why I labelled the branch with a version number. I'm going to use the same warning in the other versions.

I think I'll keep it as is.

RayOffiah: Nope. I've changed my mind. It's overkill for one  single line of text, even if's going in multiple versions.

ElliotFrancisHunter (approved):
Looks good to me! 👍🏼

PR #822, opened 2024-09-30
DOC-12595: Add missing config value

ElliotFrancisHunter:
Quick one liner for a customer found issue. Adding missing TLS version to tls_minimum version on the legacy config.

Ticket: https://jira.issues.couchbase.com/browse/DOC-12595

Preview showing successful change:
https://preview.docs-test.couchbase.com/legconfig/sync-gateway/current/configuration-properties-legacy.html#tls_minimum_version

### DIFF
--- a/modules/ROOT/assets/attachments/configuration-properties-legacy.yaml
+++ b/modules/ROOT/assets/attachments/configuration-properties-legacy.yaml
@@ -1727,5 +1727,6 @@ properties:
       - "tlsv1"
       - "tlsv1.1"
       - "tlsv1.2"
+      - "tlsv1.3"
     type: string
     default: 'tlsv1'
--- a/modules/ROOT/assets/attachments/configuration-properties-legacy.yaml
+++ b/modules/ROOT/assets/attachments/configuration-properties-legacy.yaml
@@ -1727,6 +1727,6 @@ properties:
       - "tlsv1"
       - "tlsv1.1"
       - "tlsv1.2"
-      - "tlsv1.3"
+      - "tlsv1.2"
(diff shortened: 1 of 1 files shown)


torcolvin:
This is legacy documentation, but I think it can be worthwhile to update. 

The acceptable values match https://docs.couchbase.com/sync-gateway/current/configuration-schema-bootstrap.html#lbl-schema (which autogenerated from https://docs.couchbase.com/sync-gateway/current/rest_api_admin.html#tag/Server/operation/get__config) or https://docs.couchbase.com/sync-gateway/current/configuration-schema-database.html#DatabaseConfig (which is generated from https://docs.couchbase.com/sync-gateway/current/rest_api_admin.html#tag/Database-Configuration/operation/put_db-_config).

Occasionally the default value _might_ be different, for example: database `enable_xattrs` which defaults to false in the legacy configuration but is true in non legacy configuration.

modules/ROOT/assets/attachments/configuration-properties-legacy.yaml
```diff
@@ -1727,5 +1727,6 @@ properties:
       - "tlsv1"
       - "tlsv1.1"
       - "tlsv1.2"
+      - "tlsv1.3"
     type: string
     default: 'tlsv1'
```
torcolvin:
```suggestion
    default: 'tlsv1.2'
```

iamfortune: Updated and Approving

iamfortune (approved):

PR #825, opened 2024-10-17
DOC-12649 update Metrics API version number

osfameron:


torcolvin:
The version number is correct on Sync Gateway 3.2 page.

PR #839, opened 2025-01-07
DOC-12841: Metrics REST API doc fixes

simon-dew:
Docs issue: [DOC-12841](https://jira.issues.couchbase.com/browse/DOC-12841)

This PR updates the REST API documentation for Sync Gateway _only_. Updates to the REST API documentation for App Services will follow in a separate PR.

Docs preview:

- [Sync Gateway Metrics REST API (3.2) › JSON](https://preview.docs-test.couchbase.com/DOC-12841/sync-gateway/current/rest_api_metric.html#tag/JSON)
  - [x] Updated **Get all Sync Gateway statistics** endpoint so that it appears under the **JSON** tag
  - [x] Updated summary of **Get all Sync Gateway statistics** endpoint to include **in JSON Format**
  - [x] Added link from 200 response to JSON Metrics documentation
- [Sync Gateway Metrics REST API (3.2) › Prometheus](https://preview.docs-test.couchbase.com/DOC-12841/sync-gateway/current/rest_api_metric.html#tag/Prometheus)
  - [x] Removed duplicate `_metrics` endpoint
  - [x] Added link from 200 response to Prometheus Metrics documentation
  - [x] Included minimal Prometheus metrics schema and example in 200 response
- [Metrics REST API](https://preview.docs-test.couchbase.com/DOC-12841/sync-gateway/current/rest-api-metrics.html)
  - [x] Removed broken **Use the REST API?** link from this page
  - [x] Removed Swagger UI from this page to avoid confusion
- [Public REST API](https://preview.docs-test.couchbase.com/DOC-12841/sync-gateway/current/rest-api.html)
  - [x] Removed broken **Use the REST API?** link from this page
- Enhanced all REST API reference docs with more specific titles and descriptions
- Updated all REST API reference docs to use latest 3.2.1 OpenAPI spec
- Changed `_metrics` to `metrics` throughout
- Minor fixes

Credentials: [Preview docs for internal review](https://couchbasecloud.atlassian.net/wiki/x/dYF_dQ#Preview-docs-for-internal-review)

[DOC-12841]: https://couchbasecloud.atlassian.net/browse/DOC-12841?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


simon-dew:
# Technical notes

All the updates to the Redocly REST API reference docs are driven by `generate-docs.sh`:

* Uses sparse checkout to get the OpenAPI specs from the `sync_gateway` repo
* Uses `jq` to convert the Prometheus metrics file into a JSON schema that can be consumed by Redocly
* Uses [OpenAPI overlays](https://spec.openapis.org/overlay/v1.0.0.html) to update the REST API specs for documentation output

modules/ROOT/generate-docs.sh
```diff
@@ -2,14 +2,42 @@
...
 set -exuo pipefail
 
-cd $(dirname $0)
-
 GIT_ROOT=$(git rev-parse --show-toplevel)
 
-PATH_TO_SYNC_GATEWAY=${PATH_TO_SYNC_GATEWAY:-${GIT_ROOT}/../sync_gateway/}
+# Get upstream sync_gateway repo
+cd "${GIT_ROOT}"
+branch=release/3.2.1
```
torcolvin: We create a tag for docs that is `3.2-docs` or major.minor-docs for each release. This allows us to do docs only fixes before or after release fixes.

modules/ROOT/generate-docs.sh
```diff
@@ -2,14 +2,42 @@
...
+
+[[ -d ./sync_gateway ]] || git clone --no-checkout https://github.com/couchbase/sync_gateway.git
+cd sync_gateway
+git sparse-checkout init --cone
+git sparse-checkout set docs/api
```
torcolvin, line 18: I'd be cautious about pulling a subset of directories and this repo is pretty small. This needs to have .redocly.yaml - https://github.com/couchbase/sync_gateway/blob/main/.redocly.yaml which does some processing of documentation using javascript.

modules/ROOT/generate-docs.sh
```diff
@@ -2,14 +2,42 @@
...
 
-PATH_TO_BUNDLE=${PATH_TO_BUNDLE:-${GIT_ROOT}/modules/ROOT/assets/attachments/}
+PATH_TO_BUNDLE=${PATH_TO_BUNDLE:-${GIT_ROOT}/modules/ROOT/assets/attachments}
+PATH_TO_OVERLAY=${PATH_TO_OVERLAY:-${GIT_ROOT}/modules/ROOT/assets/overlays}
+PATH_TO_TEMP=${PATH_TO_TEMP:-${GIT_ROOT}/modules/ROOT/assets/bld}
+
+# Create a bld directory to hold temporary files, if it doesn't already exist
+[[ -d "${PATH_TO_TEMP}" ]] || mkdir "${PATH_TO_TEMP}"
+
+# Use jq to transform the Prometheus metrics file into JSON schema.
```
torcolvin, line 39: I wonder if you should auto generate this file. The way to generate this would be to `cd sync_gateway/tools/stats-defintion-exporter && go run .` and this will output a current `metrics_metadata.json` in the cwd.

this might be easier than grabbing this from latestbuilds, especially because you already have the repo checked out here.

modules/ROOT/generate-docs.sh
```diff
@@ -21,14 +49,25 @@ generate() {
...
     npx @redocly/cli \
       build-docs \
       --config redocly.yaml \
-      "${PATH_TO_SYNC_GATEWAY}/docs/api/${WHAT}.yaml" \
+      "${PATH_TO_TEMP}/${WHAT}.yaml" \
       --output partials/sgw-openapi-$WHAT.html \
       --template partials/redocAll.hbs
 }
 
 generate public
```
torcolvin, line 78: I wonder if your local .redocly.yaml needs to extend the .redocly.yaml that is in sync_gateway repo? then it pulls in the code that we have to strip internal APIs.

For capella, this is even more important because it will run some javascript code to process out capella-only information.

modules/ROOT/assets/overlays/admin.yaml
```diff
@@ -0,0 +1,25 @@
+overlay: 1.0.0
+
+info:
+  title: Overlay for Admin REST API spec
+  version: 1.0.0
+
+extends: https://raw.githubusercontent.com/couchbase/sync_gateway/refs/heads/main/docs/api/admin.yaml
```
torcolvin: This gets post processed, can this use ../attachments/bundled-admin.yaml instead? This is also fixed to main which isn't necessarily going to be right. Right now main is 3.3 release.

We could also just put some changes into sync_gateway repo for description and then it only has external_docs. This might be more relevant for capella, not sure.

modules/ROOT/assets/overlays/metric.yaml
```diff
@@ -0,0 +1,94 @@
+overlay: 1.0.0
+
+info:
+  title: Overlay for Metrics REST API spec
+  version: 1.0.0
+
+extends: https://raw.githubusercontent.com/couchbase/sync_gateway/refs/heads/main/docs/api/metric.yaml
```
torcolvin: As before, can this use:

 This gets post processed, can this use ../attachments/bundled-metric.yaml instead? This is also fixed to main which isn't necessarily going to be right. Right now main is 3.3 release.

modules/ROOT/assets/overlays/public.yaml
```diff
@@ -0,0 +1,25 @@
+overlay: 1.0.0
+
+info:
+  title: Overlay for Public REST API spec
+  version: 1.0.0
+
+extends: https://raw.githubusercontent.com/couchbase/sync_gateway/refs/heads/main/docs/api/public.yaml
```
torcolvin: As before:

This gets post processed, can this use ../attachments/bundled-public.yaml instead? This is also fixed to main which isn't necessarily going to be right. Right now main is 3.3 release.

modules/ROOT/assets/overlays/admin.yaml
```diff
@@ -0,0 +1,25 @@
+overlay: 1.0.0
+
+info:
+  title: Overlay for Admin REST API spec
+  version: 1.0.0
```
torcolvin, line 5: Is this version going to show up in the display? The parent has 3.2 as the version https://github.com/couchbase/sync_gateway/blob/3.2-docs/docs/api/admin.yaml#L13

modules/ROOT/pages/rest-api-metrics.adoc
```diff
@@ -19,15 +19,15 @@ include::partial$_show_page_header_block.adoc[]
...
 It exposes Sync Gateway's stats on two endpoints, which deliver the same data in one of two formats:
 
-* The `_metrics` endpoint returns Sync Gateway statistics in {glos-term-prometheus} format. +
-For example: `GET host:4986/_metrics` +
+* The `metrics` endpoint returns Sync Gateway statistics in {glos-term-prometheus} format. +
+For example: `GET host:4986/metrics` +
 +
 For information on integration with Prometheus -- see {xref-sgw-pg-stats-prometheus} and our blog entry {url-blog-prometheus}.
 
 * The `_expvars` endpoint returns Sync Gateway statistics in JSON format.
```
torcolvin:
```suggestion
* The `_expvar` endpoint returns Sync Gateway statistics in JSON format.
```

modules/ROOT/pages/stats-monitoring-json.adoc
```diff
@@ -19,7 +19,7 @@ include::partial$_show_page_header_block.adoc[]
 include::partial$_define_stats_item_names.adoc[]
 
 == JSON format
-Use the `_expvars` method to request these metrics:
+Use the `_expvars` endpoint to request these metrics:
```
torcolvin:
```suggestion
Use the `_expvar` endpoint to request these metrics:
```

Should this be /_expvar instead ? Not sure about the style for this.

modules/ROOT/pages/stats-monitoring-prometheus.adoc
```diff
@@ -22,14 +22,14 @@ include::partial$_define_stats_item_names.adoc[]
...
-Use the `_metrics` method to request these metrics:
+Use the `metrics` endpoint to request these metrics:
 
 [#ex-metprom]
 .Metrics in Prometheus format
 ====
 [source,console]
 ----
-curl -X GET "http://localhost:4986/_metrics" -H "accept: application/json"
+curl -X GET "http://localhost:4986/metrics" -H "accept: application/json"
```
torcolvin, line 32: nit: always wondered about this for docs, should we omit `-X GET` from these? curl will do GET by default.

modules/ROOT/pages/_partials/prometheus-config.yaml
```diff
@@ -7,7 +7,7 @@ rule_files: # <.>
 
 scrape_configs:
   - job_name: swg
```
torcolvin: this is probably just a weird typo for `sgw`

modules/ROOT/generate-docs.sh
simon-dew: That's okay, `sparse-checkout` gets all the files in each directory in the path, so `.redocly.yaml` is included in the  checkout

modules/ROOT/generate-docs.sh
simon-dew: Ah, that's interesting, thanks! I was wondering how that file was generated. Currently we rely on extracting that file from a build package by hand.

For the purposes of this PR, I think we can assume that `metrics_metadata.json` exists and is up-to-date, because it is already used to generate the static Prometheus metrics documentation in [this page](https://docs.couchbase.com/sync-gateway/current/stats-monitoring-prometheus.html). But I'll talk to @osfameron about the process of generating that page and see if we can incorporate your method.

modules/ROOT/assets/overlays/admin.yaml
simon-dew: The `extends` key is just a mnemonic for users to show which file the overlay is extending, it doesn't actually do anything. The file which the overlay extends is selected by the `openapi-overlays-js`command in [generate_docs.sh](https://github.com/couchbase/docs-sync-gateway/pull/839/files#diff-327361a918a7861cbe69407fe6a73895fa27d3f5cc86d03c5b9e7183f458668b).

I can see that this line is confusing though so I'll either update it or just delete it.

modules/ROOT/assets/overlays/admin.yaml
simon-dew: It doesn't show up in the display. This is the version of the overlay, rather than the OpenAPI spec. I don't want to tie it to the version of the OpenAPI spec because it would require extra work to update the overlay files every time the spec updates, for no benefit. This key is required by the OpenAPI overlay schema, otherwise I wouldn't have bothered to include it. I'll add a note to the overlay file explaining this.

modules/ROOT/assets/overlays/admin.yaml
simon-dew: Done

modules/ROOT/assets/overlays/admin.yaml
simon-dew: Done

modules/ROOT/pages/_partials/prometheus-config.yaml
simon-dew: Fixed

modules/ROOT/pages/stats-monitoring-json.adoc
simon-dew: Good catch, fixed here and throughout

modules/ROOT/assets/overlays/metric.yaml
simon-dew: Updated as for `overlays/admin.yaml`

modules/ROOT/assets/overlays/public.yaml
simon-dew: Updated as for `overlays/admin.yaml`

modules/ROOT/pages/rest-api-metrics.adoc
simon-dew: Done

modules/ROOT/pages/stats-monitoring-prometheus.adoc
simon-dew: Good point, but it seems to be the style throughout the Sync Gateway docs (AFAICT) and I don't want to introduce inconsistency

osfameron: We're talking about a REST API call, being explicit with the HTTP method is acceptable, and provides value to people who don't have a good mental model of `curl`'s defaults

modules/ROOT/generate-docs.sh
ElliotFrancisHunter: @torcolvin Will this always be in the major.minor-docs format? If so then would there be a way to get this information from the docs antora.yaml so we can remove the risk of having to remember to manually update this each time a new major or minor version is released @osfameron?

torcolvin: It is always in major.minor-docs

modules/ROOT/generate-docs.sh
```diff
@@ -2,14 +2,42 @@
...
+# Get upstream sync_gateway repo
+cd "${GIT_ROOT}"
+branch=release/3.2.1
+
+[[ -d ./sync_gateway ]] || git clone --no-checkout https://github.com/couchbase/sync_gateway.git
+cd sync_gateway
+git sparse-checkout init --cone
+git sparse-checkout set docs/api
+git checkout ${branch}
+git pull --depth=1 origin ${branch}
```
torcolvin: Since 3.2-docs is a tag and not a branch, I think we want to do something different here. 

The change from `git pull --depth=1 origin ${branch}` to `git reset --hard ${tag}` addresses if the existing tag is not on the same branch as before. In practice, none of this will happen due to removing the sync gateway checkout.

I renamed variable from branch to tag to be more clear.

```suggestion
# check & delete local tag in case remote tag has been moved
has_tag=$(git tag -l ${tag})
if [[ -n ${has_tag} ]]; then
    git tag -d ${tag}
fi
git reset --hard ${tag}
```

modules/ROOT/assets/overlays/admin.yaml
```diff
@@ -0,0 +1,25 @@
...
+
+extends: https://raw.githubusercontent.com/couchbase/sync_gateway/refs/heads/main/docs/api/admin.yaml
+
+actions:
+  - description: Update title and introduction
+    target: $.info
+    update:
+      title: Sync Gateway Admin REST API
+      description: |-
```
torcolvin, line 17: I had been confused by this before, but I think that we want to update all the code we can in the sync gateway repo with the info + tags/descriptions, and then keep this overlay document with only:

```
  - description: Update link to external docs
    target: '$'
    update:
      externalDocs:
        description: ⬆️ Admin REST API Overview
        url: rest-api-admin.html
```

I am happy for you to do the PR against Sync Gateway for this, it would be against `main`. Then docs or SGW team can create backport to 3.2 branch, and we would move our tag.

torcolvin (commented):
I had soret

modules/ROOT/generate-docs.sh
```diff
@@ -2,14 +2,42 @@
 
 set -exuo pipefail
 
-cd $(dirname $0)
```
osfameron, line 5: This allowed us to call the script from another directory, such as docs-site. Not a biggie, but a minor convenience while testing.

modules/ROOT/generate-docs.sh
osfameron: @ElliotFrancisHunter  ah you mean we could calculate this number from the `antora.yml` instead of having to update this script? Sounds sensible.

modules/ROOT/generate-docs.sh
osfameron: At least as this is a new checkout you don't have to mess about with `git fetch --tags --force` as you have to do with an existing one 😓

modules/ROOT/generate-docs.sh
simon-dew: I think it should still work? The script now cd's to `${GIT_ROOT}` straight away in order to do the sparse checkout: see line 8. It then cd's to `modules/ROOT` to generate the documents.

modules/ROOT/generate-docs.sh
simon-dew: Done. Adds `yq` as another dependency to the script.

modules/ROOT/generate-docs.sh
simon-dew: @torcolvin the suggested change generates an error when I run the script. (You can see I'm parsing the `antora.yaml` file to set the `$tag` environment variable to `3.2-docs`.)

``` bash
++ yq -r .version antora.yml
+ tag=3.2-docs
+ [[ -d ./sync_gateway ]]
+ git clone --no-checkout https://github.com/couchbase/sync_gateway.git
Cloning into 'sync_gateway'...
remote: Enumerating objects: 72428, done.
remote: Counting objects: 100% (1198/1198), done.
remote: Compressing objects: 100% (484/484), done.
remote: Total 72428 (delta 1031), reused 722 (delta 714), pack-reused 71230 (from 3)
Receiving objects: 100% (72428/72428), 57.23 MiB | 3.53 MiB/s, done.
Resolving deltas: 100% (54428/54428), done.
+ cd sync_gateway
+ git sparse-checkout init --cone
+ git sparse-checkout set docs/api
++ git tag -l 3.2-docs
+ has_tag=3.2-docs
+ [[ -n 3.2-docs ]]
+ git tag -d 3.2-docs
Deleted tag '3.2-docs' (was 3bb180eb9)
+ git reset --hard 3.2-docs
fatal: ambiguous argument '3.2-docs': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

If I remove the local tag check  (the `has_tag` line and the `if` block) and just include the `git reset --hard ${tag}` line, the script runs perfectly.

> In practice, none of this will happen due to removing the sync gateway checkout.

Are you happy with me just removing the check?

modules/ROOT/generate-docs.sh
osfameron: I don't think so?
`GIT_ROOT=$(git rev-parse --show-toplevel)` is run from the current working directory, so will end up being wrong if you are calling from somewhere else won't it?

modules/ROOT/assets/overlays/admin.yaml
simon-dew: Okay, good to know, thanks. I'll probably look into doing this under a follow-up Jira ticket, rather than as part of this issue.

modules/ROOT/generate-docs.sh
torcolvin: The problem is that if the tag is moved by us to move to a newer version of 3.2, this will fail. I missed a critical line of git fetch --tags again after deleting the local one but before resetting to the new one.

```
git fetch --tags # update remote tags after deleting local ones
git reset --hard 3.2-docs
```

modules/ROOT/generate-docs.sh
simon-dew: Ah I understand. I'll put this line back in!

torcolvin: This has its own downsides, but this script itself could cd to the directory of the script: 

```
SCRIPT_DIR=$(dirname "${BASH_SOURCE[0]}")
cd $SCRIPT_DIR
```

which would make it robust to running from out of directory.

modules/ROOT/assets/overlays/admin.yaml
torcolvin: I think the intention might be to fix as part of https://jira.issues.couchbase.com/browse/CBG-4312 - I'd definitely consider putting that fix in before you do capella side of this.

modules/ROOT/generate-docs.sh
simon-dew: Done

modules/ROOT/generate-docs.sh
simon-dew: Done

modules/ROOT/generate-docs.sh
simon-dew: Well, we're already using the `.redocly.yaml` that is in the `sync_gateway` repo when we create the bundled spec, as the first step in the `generate()` routine. However you are right, it is important that we need to run all the filters that the upstream `.redocly.yaml` specifies when we create the bundled spec. So I've updated the bundle step to specify the API by the name assigned in the `apis` section of the upstream `.redocly.yaml`.

modules/ROOT/generate-docs.sh
```diff
@@ -2,33 +2,82 @@
...
 GIT_ROOT=$(git rev-parse --show-toplevel)
+cd "${GIT_ROOT}"
 
-PATH_TO_SYNC_GATEWAY=${PATH_TO_SYNC_GATEWAY:-${GIT_ROOT}/../sync_gateway/}
+# Get docs tag name from current major.minor version
+yq -V || brew install yq
+tag=$(yq -r .version antora.yml)-docs
+
+# Get upstream sync_gateway repo
+[[ -d ./sync_gateway ]] || git clone --no-checkout https://github.com/couchbase/sync_gateway.git
```
osfameron: Why are we only cloning conditionally?
If the directory already exists, will the `sparse-checkout` commands below work as expected?

modules/ROOT/generate-docs.sh
simon-dew: Answer from @osfameron was "no" to this suggestion:

> I'd prefer to consume developer artefacts, not generate them
> learning how to build Go dependencies and then periodically maintaining that toolchain when the upstream devs change things adds no value to docs toolchain.

modules/ROOT/generate-docs.sh
simon-dew: Good point. This line is left over after copying / pasting from the docs-analytics repo, where the sparse checkout persists after the script is run. Here, we delete the sparse checkout at the end of the script, so we can probably assume the directory doesn't already exist.

simon-dew: (The point is that cloning can take a while; the `assembly.sh` script in Analytics assumes that if the directory is already there, it's probably the cloned repo)

simon-dew: However, I've found in the Analytics repo that persisting the cloned upstream repo can lead to problems: usually git ignores the cloned upstream repo (good), but _sometimes_ the cloned upstream repo gets randomly added to commits in the Analytics repo (bad). That was why I decided in this case to delete the cloned upstream repo after the script has finished with it, and put up with the delay caused by cloning each time.

simon-dew: Done

idulo:
Thanks @simon-dew the responses of the JSON and Prometheus API look very user friendly. LGTM ✅

idulo (approved):

ElliotFrancisHunter (approved):
Happy to approve this. Big improvement! 👍🏼

PR #841, opened 2025-02-03
DOC-11438: SGW-Server bucket compat page

ElliotFrancisHunter:
# PR for ticket: https://jira.issues.couchbase.com/browse/DOC-11438
Adding copy about SGW not supporting non-default durability settings on buckets.

## Changed Pages
-------------------

- SGW/Server bucket compatibility page - https://preview.docs-test.couchbase.com/durability/sync-gateway/current/server-compatibility-buckets.html

----------------------
## Pages still needing changes/updates
-----------------------
...
-----------------------

### DIFF
--- a/modules/ROOT/pages/server-compatibility-buckets.adoc
+++ b/modules/ROOT/pages/server-compatibility-buckets.adoc
@@ -1,0 +1,42 @@
+= Buckets -- Server Compatibility
+ifdef::show_edition[:page-edition: {release}]
+ifdef::prerelease[:page-status: {prerelease}]
+:page-role:
+:description: pass:q[How _Sync Gateway_ works with _Couchbase Server's_ _Buckets_]
+:idprefix:
+:idseparator: -
+
+
+include::partial$_set_page_context.adoc[]
+
+
+// BEGIN -- Page Attributes
+// END -- Page Attributes
+
+
+// BEGIN -- Page Header
+:param-topic-group: compatibility
+:param-related: {compatibility--xref}
+include::partial$_show_page_header_block.adoc[]
+// END -- Page Header
+
+== Introduction
+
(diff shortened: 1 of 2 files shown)


osfameron (approved):

modules/ROOT/pages/server-compatibility-buckets.adoc
```diff
@@ -0,0 +1,56 @@
...
+
+TIP: See: {compatibility--xref} for version compatibility information.
+
+You can find details here about compatibility between Couchbase Server buckets and the Couchbase Mobile ecosystem.
+
+== Durability
+
+Sync Gateway does not support non-default durability settings and therefore xref:server:learn:data/durability.adoc[durable writes].
+Make sure your buckets have their bucket durability setting set to `None`.
+If this is not the case you'll get a xref:server:learn:data/durability.adoc#failure-scenarios[failure scenario], `Write while SyncWrite is pending`, and the attempt at a durable write fails.
```
bbrks, line 38: Might be worth making a note here that clients can still use the higher durability settings when set on the client side. This incompatibility is just from the bucket-wide durability.

More info here about how client and bucket level settings interact with each other:
https://docs.couchbase.com/server/current/learn/data/durability.html#specifying-levels

bbrks (commented):
Looks OK - maybe we should include the info about client-durability setting, since having durable writes is still possible and may be desirable for customers - but cannot be set at the bucket level.

modules/ROOT/pages/server-compatibility-buckets.adoc
ElliotFrancisHunter: Added a couple extra lines to clarify this.

bbrks (approved):
LGTM 👍

PR #854, opened 2025-05-01
DOC-12386: Replace {tkn-db}

simon-dew:
Docs issue: [DOC-12386](https://jira.issues.couchbase.com/browse/DOC-12386)

Starting with release/2.8, which is the earliest supporting version. Will merge this forward to all following versions, with modifications as required per version.

(EDIT: apparently 2.8 is at end-of-support, but having got this far I don't think it's worth abandoning this PR and starting again.)

Related PR for Couchbase Lite:

* https://github.com/couchbase/docs-couchbase-lite/pull/969

[DOC-12386]: https://couchbasecloud.atlassian.net/browse/DOC-12386?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


torcolvin:
Is the fix for this to include: `include::partial$_set_page_context.adoc[]` since `{tkn-db}` is used elsewhere and defined in https://github.com/couchbase/docs-sync-gateway/blob/f6d013e9c849c0c8b79397ec9e5604ed28515257/modules/ROOT/pages/_partials/_define_page_index.adoc#L27

I'm not very knowledgeable about the docs.

simon-dew:
> Is the fix for this to include: `include::partial$_set_page_context.adoc[]` since `{tkn-db}` is used elsewhere and defined in
> 
> https://github.com/couchbase/docs-sync-gateway/blob/f6d013e9c849c0c8b79397ec9e5604ed28515257/modules/ROOT/pages/_partials/_define_page_index.adoc#L27

I looked into that. In most pages, the include directives which referenced `_define_page_index` partial were commented out. And in any case, all instances of `{tkn-db}` were escaped (either with a backslash before the first brace, or with `+` signs surrounding the reference), so that even if the `_define_page_index` partial was included correctly, the `{tkn-db}` attribute would never have been replaced. Seems to me the simplest thing to do is just to replace the `{tkn-db}` with the actual path parameter as specified in the API documentation.

> I'm not very knowledgeable about the docs.

I'm not sure what the original author's intent was here either 😉

torcolvin:
> > Is the fix for this to include: `include::partial$_set_page_context.adoc[]` since `{tkn-db}` is used elsewhere and defined in
> > https://github.com/couchbase/docs-sync-gateway/blob/f6d013e9c849c0c8b79397ec9e5604ed28515257/modules/ROOT/pages/_partials/_define_page_index.adoc#L27
> 
> I looked into that. In most pages, the include directives which referenced `_define_page_index` partial were commented out. And in any case, all instances of `{tkn-db}` were escaped (either with a backslash before the first brace, or with `+` signs surrounding the reference), so that even if the `_define_page_index` partial was included correctly, the `{tkn-db}` attribute would never have been replaced. Seems to me the simplest thing to do is just to replace the `{tkn-db}` with the actual path parameter as specified in the API documentation.
> 
> > I'm not very knowledgeable about the docs.
> 
> I'm not sure what the original author's intent was here either 😉

If this is wrong, then https://docs.couchbase.com/sync-gateway/current/database-offline.html#actions is also wrong since it has backslash before the first brace. I found by looking at `{tkn-db}` in the source, this isn't the only place this is used.

I don't have a problem with this replacement, I don't know what the purpose of `{tkn-db}` or `{sgw}` is, you can see this is also wrong as a token substition here https://docs.couchbase.com/sync-gateway/current/authentication-users.html 

It seems correct to drop any of this token substitution if it doesn't work IMO.

simon-dew:
> If this is wrong, then https://docs.couchbase.com/sync-gateway/current/database-offline.html#actions is also wrong since it has backslash before the first brace. I found by looking at `{tkn-db}` in the source, this isn't the only place this is used.

Good spot! In that case the original author has escaped the brace with a backslash, and then surrounded the entire reference with plus signs, thus escaping the backslash itself 🙄 

The instance of tkn-db on that page is fixed as part of this PR, but I'll have to check for double escaping now.

> I don't have a problem with this replacement, I don't know what the purpose of `{tkn-db}` or `{sgw}` is, you can see this is also wrong as a token substition here https://docs.couchbase.com/sync-gateway/current/authentication-users.html
> 
> It seems correct to drop any of this token substitution if it doesn't work IMO.

My guess is that at the time it was written, the exact path parameter had maybe not been finalized, so the original author tried to use an attribute instead. Without going into the rights or wrongs of that decision, I think the attribute definitely needs to be replaced with plain text now.

torcolvin:
> > If this is wrong, then [docs.couchbase.com/sync-gateway/current/database-offline.html#actions](https://docs.couchbase.com/sync-gateway/current/database-offline.html#actions) is also wrong since it has backslash before the first brace. I found by looking at `{tkn-db}` in the source, this isn't the only place this is used.
> 
> Good spot! In that case the original author has escaped the brace with a backslash, and then surrounded the entire reference with plus signs, thus escaping the backslash itself 🙄
> 
> The instance of tkn-db on that page is fixed as part of this PR, but I'll have to check for double escaping now.
> 
> > I don't have a problem with this replacement, I don't know what the purpose of `{tkn-db}` or `{sgw}` is, you can see this is also wrong as a token substition here [docs.couchbase.com/sync-gateway/current/authentication-users.html](https://docs.couchbase.com/sync-gateway/current/authentication-users.html)
> > It seems correct to drop any of this token substitution if it doesn't work IMO.
> 
> My guess is that at the time it was written, the exact path parameter had maybe not been finalized, so the original author tried to use an attribute instead. Without going into the rights or wrongs of that decision, I think the attribute definitely needs to be replaced with plain text now.

SGTM, I'm OK to approve this PR as is, or wait for you replace all the locations, let me know.

osfameron (approved):
The best idea I came up with last time I looked at this, was there'd been some confusion between informal parameters 
(e.g. like "{db}" as a signal to the *reader* to replace with their database name)
with asciidoc parameters like `{tkn-db}`?

Agree this looks sensible, will approve this and CBL ones

PR #864, opened 2025-06-13
DOC-13246: Sync Gateway 3.3 Partitioned Indexes Feature

simon-dew:
Docs issue: [DOC-13246](https://jira.issues.couchbase.com/browse/DOC-13246)

Docs preview: [Partitioned Indexes](https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-13246/sync-gateway/current/index-partitions.html)
Credentials: [Preview docs for internal review](https://couchbasecloud.atlassian.net/wiki/x/dYF_dQ#Preview-docs-for-internal-review)

Other changes include: 

- Updating the Redocly REST API pages: see https://github.com/couchbase/sync_gateway/pull/7584
- Deprecating `num_index_replicas` in favor of `index.num_replicas`

[DOC-13246]: https://couchbasecloud.atlassian.net/browse/DOC-13246?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+== When to Use Partitioned Indexes
+
+Partitioned indexes are an optimization in Sync Gateway intended only for deployments using Memory-Optimized Indexes (MOI) on Couchbase Server, and only when those MOI indexes have grown too large to fit on a single index node.
+
+[IMPORTANT]
+.Recommendation
+--
+Partitioned indexes should not be used for experimentation or general performance tuning.
+Misuse can lead to index bloat, increased resource usage, and unnecessary complexity.
```
torcolvin: If the indexes are deleted, it doesn't cause index bloat.

The side effects are: 

- increased memory/cpu on n1ql nodes.
-  Possible performance degradation of making a lot of new replication calls. We have performance test for spamming Sync Gateway with new replication requests https://jira.issues.couchbase.com/browse/MB-66535?focusedCommentId=1202572 . The data is collected, but we are trying to assemble this better now. I don't think the docs should reference this CBSE, but I wanted to include it for @idulo.

It should not increase the load on Sync Gateway nodes at all.

modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+If you are using partitioned indexes, the recommended number of partitions is 8.
+
+=== Option 1: Zero Downtime (Recommended for Production)
+
+This option is ideal for production environments where uptime is critical.
+It requires temporary over-provisioning of index nodes (2&times; size) while partitioned indexes are created.
+The advantage is no downtime; the disadvantage is temporarily higher resource usage.
+
+.Procedure
+. Spin up a new Sync Gateway instance (not in LB, no import/ISGR).
```
torcolvin: This is not required, sorry this was in the original document, but I created `/{db}/_index_init` to make this "foolproof".

modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+. Run xref:rest_api_admin_static.adoc#tag/Database-Management/operation/post_db-_index_init[`POST /{db}/_index_init`] on the new instance with `num_partitions` set to the required number of partitions -- 8 recommended.
```
torcolvin: I'm not sure how much you want to cross link information, but we do this number from https://docs.couchbase.com/cloud/n1ql/n1ql-language-reference/index-partitioning.html#number-of-partitions

modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+. Wait for completion via xref:rest_api_admin_static.adoc#tag/Database-Management/operation/get_db-_index_init[`GET /{db}/_index_init`].
+
+. Run xref:rest_api_admin_static.adoc#tag/Database-Configuration/operation/post_db-_config[`POST /{db}/_config`] to use new indexes.
```
torcolvin, line 89: @idulo is it worth mentioning configuration groups here? https://docs.couchbase.com/sync-gateway/current/configuration-overview.html#lbl-config-grp

The customers who are the target for partitioned indexes are more likely to have multiple configuration groups, IMO due to large size. Although users who don't have import will not be affected.

If there are any configuration groups running on a different configuration, calling `/_post_upgrade` can delete indexes out from under a running database.

modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+. Run xref:rest_api_admin_static.adoc#tag/Server/operation/post__post_upgrade[`POST /_post_upgrade`] to remove old indexes.
+
+. Optionally rebalance index nodes using Couchbase Server REST API.
```
torcolvin: The default behavior of rebalance doesn't affect indexes. In order to opt into index rebalance, a specific setting needs to be set. https://docs.couchbase.com/server/current/learn/clusters-and-availability/rebalance.html#index-redistribution

Again, I'm not sure if the right thing is cross linking or not.

modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+=== Option 2: With Downtime
+
+This option requires no extra resources, but involves planned downtime.
+It involves lower intermittent hardware requirements.
+
+.Procedure
+. Take the database offline via xref:rest_api_admin_static.adoc#tag/Database-Configuration/operation/post_db-_config[`POST /{db}/_config`] with `offline` set to `true`.
+
+. Manually delete old indexes -- those with names matching `^sg_.*x1$`.
```
torcolvin: This is explicitly wrong, this will delete all indexes.

I am actually leaning toward not documenting this behavior because I think it's unlikely and I do not like users knowing the details of our indexes names.

If you are running a test environment, you can probably start from scratch without any indexes.

@idulo what do you think about removing this? If we do want to keep this we want to remove the indexes for each collection attached to a database. The names of the indexes are internal to sync gateway and we do reserve the right to change them. As of Sync Gateway 3.0, if you are using `enable_shared_bucket_access=true`, they will be named `sg_allDocs_x1` and `sg_channels_x1` on each collection but I want to stress this is very much an internal API.

modules/ROOT/pages/index-partitions.adoc
```diff
@@ -0,0 +1,153 @@
...
+
+=== Identify Index Configuration
+
+To identify the current index configuration, you can:
+
+* Query xref:rest_api_admin_static.adoc#tag/Database-Management/operation/get_db-_index_init[`GET /{db}/_index_init`] to verify the `settings.num_partitions` in use.
+
+* Use xref:rest_api_admin_static.adoc#tag/Database-Configuration/operation/get_db-_config[`GET /{db}/_config`] to check the active configuration.
+
+* List existing GSI indexes on Couchbase Server via UI or REST API and identify any ending in `x1` -- the default for unpartitioned indexes.
```
torcolvin: @idulo I don't know if I feel great about including this. There will be indexes on:

- the _default._default collection always has indexes on user & roles & sync documents (and the name of this index is going to change from 3.2 to 3.3). If you upgrade from 3.2 you would have different indexes than if you create fresh on 3.3
- for each collection we have:
	- access, roleAccess, tombstones which are non partitioned always
	- channels, allDocs which optionally could be partitioned
	
It is very tricky to identify unused indexes because you have to consider all databases in a bucket and each configuration for them, which is why we pushed this logic to `/_post_upgrade` to iterate over all the databases and find actually unused data.

torcolvin (commented):
I didn't do a full review of this, I am going to come back to this. I wanted to get this in a position for PM review.

The document was correct for how I documented the behavior, but unfortunately some of this was wrong from the planned implementation and not the final implementation. I tried to make it simpler when I realized how complicated it can get.

Some of the comments are specific to dev discussion with PM and might be settled off this PR. I wanted to flag them here for reference and something that should be addressed, but the how of addressing might take more conversation.

modules/ROOT/pages/index-partitions.adoc
idulo: @torcolvin Thanks for this, so basically we should not advise users at all to look at their existing indexes to identify any unpartitioned ones, might not be very actionable/not safe to action. Let's remove this @simon-dew

modules/ROOT/pages/index-partitions.adoc
idulo: I understand where you are coming from @torcolvin What I want to communicate to users is that they do have the option of zero downtime, especially in less risky env, same but without exposing any prefixes or suffixes that our indexes have. What happens if users delete indexes that are NOT partitioned or CAN'T be partitioned such as the `access/roleAccess` indexes?  

```This option requires no extra resources, but involves planned downtime. It involves lower intermittent hardware requirements.

Procedure
Take the database offline via [POST /{db}/_config](https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-13246/sync-gateway/current/rest_api_admin.html#tag/Database-Configuration/operation/post_db-_config) with offline set to true.

Manually delete old indexes .

Bring the database online using [POST /{db}/_config](https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-13246/sync-gateway/current/rest_api_admin.html#tag/Database-Configuration/operation/post_db-_config) with index.num_partitions set to the required number of partitions.

Run integration and performance testing.

idulo: Should we instead have basically a warning saying you do have the option for zero downtime which is to manually delete non-partitioned indexes that have already been partitioned. However, check if they have been partitioned before deletion, be very cautious about what you are deleting, because you can leave your system in a messy state?

torcolvin: This is not all the indexes that need to be deleted, and I really think figuring out which indexes to be deleted is really hard. 

We could say all Sync Gateway indexes, given that it requires downtime. That is indexes starting with `sg_` but that might still effect other databases that are running, and so I don't really feel good about advising this strategy at all.

If they delete all sync gateway indexes, then they have to pay the cost of rebuilding them, but maybe this isn't worth covering at all in this document, and this isn't a common use case?

modules/ROOT/pages/index-partitions.adoc
idulo: What I am thinking for the performance tests outcome to be documented:

> Query performance is sensitive to channel design. Systems with many small channels tend to perform worse than those using fewer, larger channels. For better scalability and efficiency, prefer broader channel groupings over fine-grained ones.

@torcolvin thoughts?

idulo: OK, so let's remove the index bloat from that sentence and then document the above perf implications.

modules/ROOT/pages/index-partitions.adoc
idulo: You mean spinning up a new SG instance?

modules/ROOT/pages/index-partitions.adoc
idulo: Yes, let's mention them. So @torcolvin our recommendation is:

If you are using configuration groups (link to config groups) running on a different index configuration, calling /_post_upgrade can delete indexes out from under a running database. Please ensure that all configuration groups have consistent index configuration.

modules/ROOT/pages/index-partitions.adoc
idulo: We should remove the recommendation, don't think it is necessary to cross reference here as it is basically not explained exactly why they use it on their side.

modules/ROOT/pages/index-partitions.adoc
idulo: @simon-dew thoughts?

modules/ROOT/pages/index-partitions.adoc
idulo: Should we instead of the current Zero Downtime section, replace with:

> For production environments where uptime is critical, Sync Gateway supports a zero-downtime path to enable index partitioning. This requires temporary overprovisioning, typically doubling index node capacity, while new partitioned indexes are created alongside existing ones.

> Incorrect handling (e.g. manual index deletion) can cause data loss or leave the system in an unstable state. Incorrect deletion of internal indexes can lead to system instability and requires full index rebuilds. We strongly advise using the standard procedure with temporary downtime unless you are fully aware of the implications.

@torcolvin something along these lines?

modules/ROOT/pages/index-partitions.adoc
idulo: This is what we have for perf implications @adamcfraser

modules/ROOT/pages/index-partitions.adoc
simon-dew: Presumably this means that "(not in LB, no import/ISGR)" is not required? I've updated this step accordingly

modules/ROOT/pages/index-partitions.adoc
simon-dew: Removed the recommendation.

modules/ROOT/pages/index-partitions.adoc
simon-dew: Done

modules/ROOT/pages/index-partitions.adoc
simon-dew: Removed

modules/ROOT/pages/index-partitions.adoc
simon-dew: I've added a link to the server page.

modules/ROOT/pages/index-partitions.adoc
simon-dew: > We strongly advise using the standard procedure with temporary downtime unless you are fully aware of the implications.

I'm confused by this conversation, because this is the section on migrating _with_ downtime. The standard procedure — the one which is recommended for production — is the zero downtime procedure, with the automatic removal of old indexes.

For the moment I've updated this to say, "Manually delete old Sync Gateway indexes", and added a warning as outlined above, but changed to say "the standard procedure with zero downtime".

modules/ROOT/pages/index-partitions.adoc
simon-dew: Done

modules/ROOT/pages/index-partitions.adoc
simon-dew: Merging this PR as is, so that I can get all the changes for SGW 3.3 visible in one place, as requested. We can continue this conversation at https://github.com/couchbase/docs-sync-gateway/pull/877.

PR #914, opened 2025-08-28
Upgrade to sgw 4.0

iamfortune:
Docs Preview: [URL](https://preview.docs-test.couchbase.com/docs-sync-gateway-upgrade-to-sgw-4.0/sync-gateway/current/upgrading.html)

You will need the [Docs Team credentials](https://confluence.issues.couchbase.com/wiki/spaces/DOCS/pages/1971224949/Docs+Team+demo+site+passwords).

### DIFF
--- a/modules/ROOT/pages/upgrading.adoc
+++ b/modules/ROOT/pages/upgrading.adoc
@@ -67,12 +67,12 @@ NOTE: As with all major version upgrades, downgrading from 4.0 to earlier versio
 
 ==== Process
 
-.Step 1
+.Prerequisites
 
 . Upgrade Couchbase Server cluster to 7.6.0 or later and verify all server nodes are operational.
-. Test Sync Gateway connectivity with existing version to ensure compatibility.
+. Test Sync Gateway connectivity with existing version to verify compatibility.
 
-.Step 2
+.Step 1
 
 . Remove a node from the load balancer or stop traffic to the node.
 . Update the configuration file, if using unsupported settings:
@@ -82,7 +82,7 @@ NOTE: As with all major version upgrades, downgrading from 4.0 to earlier versio
 . Start the Sync Gateway service.
 . Verify node health via xref:rest-api:rest_api_public.adoc#tag/Server/operation/get_-[REST API].
 
-.Step 3
+.Step 2
 
 . Verify all databases are online.
 . Re-add the node to the load balancer or resume traffic to the node.
@@ -207,7 +207,16 @@ Depending on the one you choose, there may be additional consideration to keep i
 |High - duplicate entire cluster
 |Using an XDCR (Cross Data Center Replication) approach incurs some Sync Gateway downtime, but less than other approaches where Sync Gateway is shutdown during the entire Couchbase Server upgrade.
 
-NOTE: The XDCR replication must be a *one way* replication from the existing (source) Couchbase Server cluster to the new (target) Couchbase Server cluster.
+|Offline Upgrade
+|During entire upgrade
+|None
+|
+* Take Sync Gateway offline
+* Upgrade Couchbase Server using any of the options mentioned in the xref:server:install:upgrade.adoc[Upgrading Couchbase Server] documentation.
+* Bring Sync Gateway online
+|===
+
+The XDCR replication must be a *one way* replication from the existing (source) Couchbase Server cluster to the new (target) Couchbase Server cluster.
 No other writes can happen on the new (target) Couchbase Server cluster other than the writes from the XDCR replication.
 Do not configure any Sync Gateway instances to use the new (target) Couchbase Server cluster until the last step in the process.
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
 A rolling upgrade is the recommended method to upgrade a Sync Gateway cluster.
 At a high level, a rolling upgrade consists of the following steps:
 
-. The load balancer configuration is updated to stop any HTTP traffic going to the node that will be upgraded.
+. The load balancer configuration stops any HTTP traffic going to the node that's upgraded.
```
bbrks: This is changing the meaning of the sentence. It need to stop traffic to nodes that haven't been upgraded yet but soon will be.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
 
-. The upgrade is performed on the given node
+. The upgrade performs on the given node
 
-. The load balancer configuration is updated to re-balance the HTTP traffic across all nodes.
+. The load balancer configuration updates to re-balance the HTTP traffic across all nodes.
```
bbrks: I am not sure this change makes sense from a grammatical POV but would like another doc writer's input on this.

This seems to be changing from passive voice to something else and it doesn't read very well to me - but I am not the expert writer.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
 A rolling upgrade is the recommended method to upgrade a Sync Gateway cluster.
 At a high level, a rolling upgrade consists of the following steps:
 
-. The load balancer configuration is updated to stop any HTTP traffic going to the node that will be upgraded.
+. The load balancer configuration stops any HTTP traffic going to the node that's upgraded.
 
-. The upgrade is performed on the given node
+. The upgrade performs on the given node
```
bbrks: `the upgrade performs` doesn't make sense?

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+. `shared_bucket_access=false` is no longer supported - databases do not start with `shared_bucket_access=false`.
+. `allow_conflicts=true` no longer supported - databases do not start with this setting.
+
+
+.One Way Upgrade
+[CAUTION]
+--
+The migration to 4.x configuration is a ONE WAY process.
+You cannot downgrade to previous versions after upgrading to 4.0.
+To continue using legacy-mode configuration see: <<lbl-3-0-upgrade-opt-out>>
```
bbrks: This is mostly from the 2.x -> 3.0 persistent config upgrade process which overwrote old config files with new formats, and IMO doesn't really make sense to document in this way any more.

Upgrades to major/minor versions are _always_ a one way process since we don't support downgrade.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+--
+
+==== Considerations
+
+Before commencing your upgrade you should:
+
+. Verify your Couchbase Server version.
+. Remove or update `shared_bucket_access=false`, `allow_conflicts=true settings` and document any custom conflict resolution strategies.
```
bbrks: It might make sense to make these changes as part of the upgrade, not beforehand? But this _would_ work. It just adds more steps.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+. Create configuration backups and document your current replication settings.
+
+==== Process
+
+.Step 1
+
+. Upgrade Couchbase Server cluster to 7.6.0 or later and verify all server nodes are operational.
+. Test Sync Gateway connectivity with existing version and start the Sync Gateway service.
+. Update Sync Gateway configuration file to match the new server version and features.
```
bbrks: If SG is started in the bullet above, what is there to change? AFAIK we don't require any config updates from a server upgrade.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+. Remove the node from a load balancer or stop traffic to the node.
+. Install Sync Gateway 4.0 and update your configuration file.
+. Start the Sync Gateway service.
+. Verify node health via xref:rest-api:rest_api_public.adoc#tag/Server/operation/get_-[REST API].
+
+.Step 3
+
+. Verify all databases are online.
+. Test mobile client synchronization.
+. Validate Inter-Sync Gateway replication if configured.
```
bbrks: Doing this usually requires the node be in the load balancer in the first place to be accessible from a device.

This seems a little bit over-kill for an upgrade process?

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+It's the default operational mode when starting {sgw}.
 
 The feature provides a smooth upgrade path for existing users by automatically converting their existing configuration files to the new persistent configuration format.
 
 .One Way Upgrade
 [CAUTION]
 --
-The migration to 3.x configuration is a ONE WAY process; to continue using legacy-mode configuration see: <<lbl-3-0-upgrade-opt-out>>
+The migration to 3.x configuration is a ONE WAY process.
+You cannot downgrade to previous versions after upgrading to 4.0.
```
bbrks: This is in the 3.1 upgrade heading but mentions 4.0?

modules/ROOT/pages/upgrading.adoc
```diff
@@ -234,32 +284,33 @@ If the view request exceeds the default timeout of 75s (which would be expected
...
 
-Removal of the obsolete design documents is done via a call to the xref:rest-api:rest_api_admin.adoc#tag/Server/operation/post__post_upgrade[`_post_upgrade`] endpoint in Sync Gateway's Admin REST API.
+To remove the obsolete design documents, call the xref:rest-api:rest_api_admin.adoc#tag/Server/operation/post__post_upgrade[`_post_upgrade`] endpoint in Sync Gateway's Admin REST API.
 This endpoint can be run in preview mode (`?preview=true`) to see which design documents would be removed.
 
 
 == 1.x to 1.5
 
 From: *Sync Gateway 1.1, 1.2, 1.3, 1.4* +
 To:  *Sync Gateway 1.5* +
```
bbrks: Maybe track in a separate ticket initiative - but these versions are over 8 years old and really need removing from the docs.
Keeping them around introduces unnecessary maintenance burden like these changes below.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,105 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
 
 Those steps are then repeated for each node in the {sgw} cluster.
 
 
+== Upgrade to 4.0
+
+=== Upgrade Requirements
+
+. Couchbase Server 7.6.0 or later - Sync Gateway 4.0 does not support older server versions.
+. Couchbase Server 7.6.5 or later for bi-directional XDCR features.
```
bbrks: Would it be simpler to only recommend `7.6.5`?

I can't think of a good reason why customers wouldn't be on the latest maintenance version and specifying 7.6.5 as the minimum covers all technical requirements for Sync Gateway.

modules/ROOT/pages/upgrading.adoc
iamfortune: Good catch! I'll change it to: 'The load balancer configuration stops any HTTP traffic to the node you're upgrading.' This makes it clear the traffic stops during the upgrade process, not after.

modules/ROOT/pages/upgrading.adoc
iamfortune: Good catch on the grammar issues. I've revised this to: 'The upgrade runs on the given node' and 'The load balancer re-balances HTTP traffic across all nodes.' This will maintain consistency and proper voice throughout the procedure.

modules/ROOT/pages/upgrading.adoc
iamfortune: Thanks for catching that, I've updated this to reflect that.

modules/ROOT/pages/upgrading.adoc
iamfortune: Done - integrated the config changes into the upgrade steps. Thanks for the suggestion to streamline it.

modules/ROOT/pages/upgrading.adoc
iamfortune: I removed the config section since no changes are needed for server update.

modules/ROOT/pages/upgrading.adoc
iamfortune: I've simplified to just verifying databases are online before returning the node to rotation.

modules/ROOT/pages/upgrading.adoc
iamfortune: Yeah, I've taken out 1.x to 1.5x related docs, do we still need the "Migrating from Bucket Shadowing" of the Upgrade page from an engineering perspective?

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,89 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+==== Process
+
+.Step 1
+
+. Upgrade Couchbase Server cluster to 7.6.0 or later and verify all server nodes are operational.
+. Test Sync Gateway connectivity with existing version to ensure compatibility.
+
+.Step 2
+
+. Remove the node from a load balancer or stop traffic to the node.
```
bbrks: 1 to many SG nodes and 1 load balancer - the wording should be this way around:

```suggestion
. Remove a node from the load balancer or stop traffic to the node.
```

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,89 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+. Update the configuration file:
```
bbrks: These settings have been deprecated and non-default for a while now, so most customers should've already moved away from them.

```suggestion
. Update the configuration file, if using unsupported settings:
```

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,89 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+- Remove `shared_bucket_access=false`.
+- Remove `allow_conflicts=true`.
+. Install Sync Gateway 4.0 and update your configuration file.
```
bbrks: Config update was done in bulletpoint above (if any changes were required)
```suggestion
. Install Sync Gateway 4.0
```

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,89 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
+. `allow_conflicts=true` no longer supported - databases do not start with this setting.
+
+NOTE: As with all major version upgrades, downgrading from 4.0 to earlier versions is not supported.
+
+==== Process
+
+.Step 1
+
+. Upgrade Couchbase Server cluster to 7.6.0 or later and verify all server nodes are operational.
+. Test Sync Gateway connectivity with existing version to ensure compatibility.
```
bbrks: This is already part of upgrade requirements, so maybe this step should be renamed to Prerequisites to make it super-clear it should be done **before** SG upgrade starts? Customers should be able to do the server upgrade at any time before SG upgrade starts - even months before.

modules/ROOT/pages/upgrading.adoc
```diff
@@ -46,55 +46,89 @@ include::ROOT:partial$_show_page_header_block.adoc[]
...
 At a high level, a rolling upgrade consists of the following steps:
 
-. The load balancer configuration is updated to stop any HTTP traffic going to the node that will be upgraded.
+. The load balancer configuration stops any HTTP traffic to the node you're upgrading.
 
-. The upgrade is performed on the given node
+. The upgrade runs on the given node.
 
-. The load balancer configuration is updated to re-balance the HTTP traffic across all nodes.
+. The load balancer re-balances HTTP traffic across all nodes.
```
bbrks:
```suggestion
. The load balancer configuration resumes traffic to the upgraded node.
```

modules/ROOT/pages/upgrading.adoc
bbrks: Nope, shadow buckets are also very old feature and no longer present in the product.

bbrks (changes requested):
A few more tweaks based on latest changes but I think it's mostly good 👍

modules/ROOT/pages/upgrading.adoc
iamfortune: Resolved and updated in latest commit.

bbrks (approved):
LGTM 👍 thanks!

PR #918, opened 2025-09-19
whats new page for sgw 4.0

iamfortune:
Preview: [Whats New in 4.0](https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-13565-whats-new-sge-4.0/sync-gateway/current/whatsnew.html)

### DIFF
--- a/modules/ROOT/pages/whatsnew.adoc
+++ b/modules/ROOT/pages/whatsnew.adoc
@@ -20,19 +20,15 @@ include::ROOT:partial$block-caveats.adoc[tags=caution-eventing]
 
 == Overview of Sync Gateway {release}
 
-_Sync Gateway {release} represents a major architectural evolution, introducing version vectors as the foundation for document revision tracking and enabling bidirectional replication with Cross Data Center Replication (XDCR) between mobile clusters._
+Sync Gateway {release} represents a major architectural evolution, transitioning from revision trees to xref:server-compatibility:server-compatibility-xdcr-mobile.adoc#version-vectors[version vectors] as the foundation for document revision tracking.
 
-Another significant enhancement is the transition from revision trees to version vectors enabling true active-active mobile cluster deployments with Cross Data Center Replication (XDCR) while maintaining data consistency and conflict resolution capabilities.
+This architectural change enables true active-active mobile cluster deployments with bidirectional Cross Data Center Replication (XDCR) while maintaining data consistency and conflict resolution capabilities.
 
 == Cross Data Center Replication (XDCR) Interoperability
 
 * *Bi-directional XDCR between mobile clusters*
 Sync Gateway now supports two way active-active replication using Couchbase Server’s XDCR across mobile clusters.
 This enables active-active deployments with high availability and no downtime during fail over or disaster recovery.
-
-* *Unidirectional XDCR to read-only clusters*
-Simplified support for active–standby and migration scenarios.
-Sync Gateway automatically enforces read-only behavior on target clusters, removing the need for manual configuration.
 
 * *Unified replication of mobile and non-mobile documents*
 Applications can now replicate both mobile and non-mobile data using a single replication protocol,
@@ -42,8 +38,8 @@ For more information, see xref:ROOT:server-compatibility-xdcr.adoc[XDCR -- Serve
 
 == Conflict Resolution with Version Vectors
 
-* Sync Gateway 4.0 leverages *version vector–based conflict resolution* for XDCR.
-* By default, conflicts are handled using Couchbase Server’s *custom conflict resolution (CCR)*.
+* Sync Gateway 4.0 leverages version vector–based conflict resolution for XDCR.
+* By default, Sync Gateway resolves conflicts using `Last Write Wins` (LWW), with the option to use Couchbase Server's custom conflict resolution (CCR).
 * Developers can extend with custom conflict resolvers to meet application-specific needs.
 
 This ensures consistency across mobile and server data while maintaining performance during high-volume replication.
--- a/modules/ROOT/pages/whatsnew.adoc
+++ b/modules/ROOT/pages/whatsnew.adoc
@@ -16,21 +16,44 @@ include::ROOT:partial$_set_page_context.adoc[]
 :param-related: {xref-sgw-pg-release-notes}
 include::ROOT:partial$_show_page_header_block.adoc[]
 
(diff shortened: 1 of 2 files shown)


modules/ROOT/pages/whatsnew.adoc
```diff
@@ -16,29 +16,52 @@ include::ROOT:partial$_set_page_context.adoc[]
...
-Do not deploy Eventing/Sync Gateway until all SGW nodes are at version 3.2+.
-For earlier Sync Gateway versions that do not write import XATTRs, Eventing functions may encounter infinite recursions and duplicate mutations if deployed in a mixed mode SGW environment.
-This occurs if you have a mixed installation consisting of Sync Gateway 3.2 and an earlier version, and you deploy a new Eventing or Sync Gateway function during an upgrade.
+== Overview of Sync Gateway {release}
 
-== Release 4.0.0 (Q1 2025)
+_Sync Gateway {release} represents a major architectural evolution, introducing version vectors as the foundation for document revision tracking and enabling bidirectional replication with Cross Data Center Replication (XDCR) between mobile clusters._
 
-=== Version Vectors
+Another significant enhancement is the transition from revision trees to version vectors enabling true active-active mobile cluster deployments with Cross Data Center Replication (XDCR) while maintaining data consistency and conflict resolution capabilities.
```
bbrks: These two paragraphs are saying the same thing - I'd collapse into one paragraph and de-duplicate the repetitive bits (version vectors, active-active, etc.)

modules/ROOT/pages/whatsnew.adoc
```diff
@@ -16,29 +16,52 @@ include::ROOT:partial$_set_page_context.adoc[]
...
 
-=== Mobile XDCR Coexistence
+* *Bi-directional XDCR between mobile clusters*
+Sync Gateway now supports two way active-active replication using Couchbase Server’s XDCR across mobile clusters.
+This enables active-active deployments with high availability and no downtime during fail over or disaster recovery.
 
-NOTE: TODO
+* *Unidirectional XDCR to read-only clusters*
+Simplified support for active–standby and migration scenarios.
+Sync Gateway automatically enforces read-only behavior on target clusters, removing the need for manual configuration.
```
bbrks: This isn't true - SG has no read-only mode. Administrators are still required to configure a sync function that rejects any incoming document updates.

This unidirectional read-only limitation/use-case mostly goes away with the 4.0 XDCR enhancements.

modules/ROOT/pages/whatsnew.adoc
```diff
@@ -16,29 +16,52 @@ include::ROOT:partial$_set_page_context.adoc[]
...
+
+* *Unified replication of mobile and non-mobile documents*
+Applications can now replicate both mobile and non-mobile data using a single replication protocol,
+reducing complexity and conflict resolution overhead.
```
bbrks: Not sure what this is referencing. There's still mobile metadata, a mobile replication protocol used for CBL and ISGR (where relevant)

Even though XDCR can replicate all data in an active-active setup - there's still use-cases that require a subset of data (e.g. by channels) to be replicated - ISGR is still useful for this.

modules/ROOT/pages/whatsnew.adoc
```diff
@@ -16,29 +16,52 @@ include::ROOT:partial$_set_page_context.adoc[]
...
+
+For more information, see xref:ROOT:server-compatibility-xdcr.adoc[XDCR -- Server Compatibility] and xref:ROOT:xdcr-bidirectional.adoc[Bi-directional XDCR Between Mobile Clusters].
+
+== Conflict Resolution with Version Vectors
+
+* Sync Gateway 4.0 leverages *version vector–based conflict resolution* for XDCR.
+* By default, conflicts are handled using Couchbase Server’s *custom conflict resolution (CCR)*.
```
bbrks: By default XDCR uses LWW (Last Write Wins) - but has the option of CCR.

bbrks (changes requested):
Quite a few corrections needed here - let me know if you need any more info.

modules/ROOT/pages/whatsnew.adoc
iamfortune: Good catch! I've consolidated those two paragraphs to remove the repetitive mentions of version vectors, active-active, and XDCR.

modules/ROOT/pages/whatsnew.adoc
iamfortune: Thanks for this. You're right about the SG read-only capabilities, and good point about this becoming less relevant with the 4.0 XDCR enhancements. I've removed that bullet point.

modules/ROOT/pages/whatsnew.adoc
iamfortune: What would be the accurate way to describe the XDCR improvement regarding mobile vs non-mobile document handling?

modules/ROOT/pages/whatsnew.adoc
iamfortune: Thank you for pointing this out, I've updated the sentence to reflect this.

modules/ROOT/pages/whatsnew.adoc
```diff
@@ -16,29 +16,48 @@ include::ROOT:partial$_set_page_context.adoc[]
...
 
-=== Mobile XDCR Coexistence
+* *Bi-directional XDCR between mobile clusters*
+Sync Gateway now supports two way active-active replication using Couchbase Server’s XDCR across mobile clusters.
+This enables active-active deployments with high availability and no downtime during fail over or disaster recovery.
 
-NOTE: TODO
+* *Unified replication of mobile and non-mobile documents*
+Applications can now replicate both mobile and non-mobile data using a single replication protocol,
+reducing complexity and conflict resolution overhead.
```
bbrks: I think I'd forget about the replication protocol and just focus on the version unification. There are still two replication protocols between XDCR and Mobile.

```suggestion
* *Unified versioning of documents compatible with both XDCR and Mobile products*.
Allows for new deployment architectures and consistent conflict resolution.
```

modules/ROOT/pages/whatsnew.adoc
iamfortune: Cc: @bbrks

bbrks: addressed in last suggestion

bbrks (approved):
LGTM 👍

PR #921, opened 2025-09-30
4.0.0: Add CBG-4887 to release note known issues

bbrks:
This was something that came up today that we need to document as a known issue (already in Jira)


iamfortune (approved):
lgtm

PR #923, opened 2025-10-10
Update 4.0 REST API

osfameron:
Running locally can cause npm to hang while executing `npx` commands.

At a guess, it's outputting a prompt "Install module?" which is being redirected, and so remains unseen.

This commit adds explicit `--yes` to run each of these without confirmation. (It might be more elegant to set `CI=true` for the script?)


iamfortune (approved):
LGTM and yeah ci might work for this.

PR #929, opened 2025-10-22
update sgw cbs compat table

iamfortune:
PR to update SGW compatibility matrix with CBS 8.0.0

Preview: [URL](https://preview.docs-test.couchbase.com/docs-sync-gateway-add-server-8.0/sync-gateway/current/product-notes/compatibility.html)

### DIFF
--- a/modules/ROOT/partials/sgw-svr-compatibility.adoc
+++ b/modules/ROOT/partials/sgw-svr-compatibility.adoc
@@ -82,14 +82,27 @@ Compatibility Matrix
 | image:ROOT:no.png[]
 | image:ROOT:no.png[]
 
-| 3.3.0
-|
+| 4.0.0
+| Bidirectional Active-Active XDCR
+| image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
 | image:ROOT:no.png[]
 | image:ROOT:no.png[]
 | image:ROOT:no.png[]
 | image:ROOT:no.png[]
 | image:ROOT:no.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
+
+| 3.3.0
+|
+| image:ROOT:yes.png[]
+| image:ROOT:yes.png[]
+| image:ROOT:yes.png[]
+| image:ROOT:yes.png[]
+| image:ROOT:yes.png[]
+| image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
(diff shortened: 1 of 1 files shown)


modules/ROOT/partials/sgw-svr-compatibility.adoc
```diff
@@ -57,29 +57,35 @@ include::ROOT:partial$block-caveats.adoc[tags="cbs6.0ke-xattrs"]
...
 Compatibility Matrix
-[cols="^1,3,^1,^1,^1,^1,^1,^1,^1,^1"]
+[cols="^1,3,^1,^1,^1,^1,^1,^1,^1,^1,^1,^1"]
 |===
 
 2+^.>|Sync Gateway ↓
-8+^|Couchbase Server →
+10+^|Couchbase Server →
 ^.>| Version ^.>| Scenario
+^.>| 8.0.0
```
gregns1, line 66: Bit of a nit pick but I'd change this to be just 8.0

modules/ROOT/partials/sgw-svr-compatibility.adoc
```diff
@@ -57,29 +57,35 @@ include::ROOT:partial$block-caveats.adoc[tags="cbs6.0ke-xattrs"]
...
 | image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
 | image:ROOT:yes.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
+| image:ROOT:no.png[]
```
gregns1, line 84: I would add a new row below 4.0.0 Sync Gateway version for 4.0.0 again but with `Bidirectional Active-Active XDCR` in the scenario column and then have 8.0 and 7.6.5 ticked as compatible with rest not compatible if that makes sense?

gregns1 (changes requested):
Couple of comments plus:

Image

The Sync Gateway versions of 3.3.0, 3.2.0, 3.1.0 and 3.0.3 all seem to have Server versions 7.6.5, 7.6.4, 7.6, 7.2 (and 7.1 for 3.3.0) as not compatible. I'm not aware of any restriction as to why they wouldn’t be compatible so these should all have tickets under them

modules/ROOT/partials/sgw-svr-compatibility.adoc
iamfortune: We have to follow the same pattern with the rest of the page, for now might have to leave it as is.

gregns1: That's fine 😄  (it was just me being picky anyway)

modules/ROOT/partials/sgw-svr-compatibility.adoc
iamfortune: You're right, I just made the changes in the most recent commit and updated the preview URL

gregns1 (approved):

PR #972, opened 2026-05-27
DOC-14380-cluster-compat-upgrade-restructure

iamfortune:
Docs Issue: [DOC-14380](https://jira.issues.couchbase.com/browse/DOC-14380?atlOrigin=eyJpIjoiY2U2MTc4MWJmODRiNDQ2Njg5ZmI3MzVmZGZiNGFiNzMiLCJwIjoiaiJ9)

PR to restructure the Upgrade page, add rolling updates and cluster compatibility page

Preview URL:
https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-14380-cluster-compat-upgrade

[DOC-14380]: https://couchbasecloud.atlassian.net/browse/DOC-14380?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/ROOT/pages/cluster-compatibility-version.adoc
+++ b/modules/ROOT/pages/cluster-compatibility-version.adoc
@@ -17,7 +17,7 @@ include::ROOT:partial$_show_page_header_block.adoc[]
 
 The cluster compatibility version is a cluster-wide value that {sgw} uses to coordinate behaviour across nodes running different versions during a rolling upgrade.
 
-Introduced in Sync Gateway 4.1, the cluster compatibility version lets you:
+Introduced in {sgw} 4.1, the cluster compatibility version lets you:
 
 * Perform a rolling upgrade without downtime or reduction in cluster capacity.
 * Preserve a safe rollback path during the upgrade window.
@@ -34,7 +34,8 @@ As you upgrade nodes one by one, the minimum advances when older nodes are repla
 
 {sgw} uses the compatibility version to gate cluster-wide feature activation.
 Features that could cause incompatibilities with older nodes remain inactive until the compatibility version reaches the version that introduced them.
-New features that do not depend on metadata shared by other Sync Gateway nodes activate immediately on each upgraded node regardless of the compatibility version. These are considered 'node-local' features, and include areas such as caching, read-only operations, general bug fixes and performance improvements.
+New features that do not depend on metadata shared by other {sgw} nodes activate immediately on each upgraded node regardless of the compatibility version.
+These are considered node-local features, and include areas such as caching, read-only operations, general bug fixes, and performance improvements.
 
 On a homogeneous cluster where all nodes run the same version, the compatibility version equals that version.
 The value is not present in the API response immediately after startup before any node has registered.
@@ -66,7 +67,7 @@ Use this procedure when you want to preserve the option to roll back during the 
 
 === Prerequisites
 
-* All nodes are running {sgw} 4.0 or later.
+* All nodes are running {sgw} 4.1 or later.
 * You have Admin REST API access to the cluster.
 * See xref:ROOT:rolling-upgrade.adoc[Rolling Upgrade] for version requirements and the node upgrade procedure.
 
@@ -81,16 +82,16 @@ Call xref:ROOT:rest_api_admin_static.adoc#tag/Server/operation/get__cluster_comp
 GET /_cluster_compat_version
 ----
 
-Example response on a homogeneous 4.0 cluster:
+Example response on a homogeneous 4.1 cluster:
 
 [source,json]
 ----
 {
-  "cluster_compat_version": "4.0",
+  "cluster_compat_version": "4.1",
   "nodes": {
-    "node1-uid": "4.0",
-    "node2-uid": "4.0",
-    "node3-uid": "4.0"
+    "node1-uid": "4.1",
+    "node2-uid": "4.1",
+    "node3-uid": "4.1"
   }
 }
 ----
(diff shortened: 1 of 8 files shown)


modules/ROOT/pages/cluster-compatibility-version.adoc
```diff
@@ -0,0 +1,221 @@
...
+The cluster compatibility version is a cluster-wide value that {sgw} uses to coordinate behaviour across nodes running different versions during a rolling upgrade.
+
+Introduced in Sync Gateway 4.1, the cluster compatibility version lets you:
+
+* Perform a rolling upgrade without downtime or reduction in cluster capacity.
+* Preserve a safe rollback path during the upgrade window.
+* Ensure that new cluster-wide features activate only after all nodes reach the new version.
+
+NOTE: Downgrading after the full cluster has been upgraded is not supported.
+The cluster compatibility version mechanism provides a rollback path only while at least one node is still on the previous version.
```
bbrks:
```suggestion
The cluster compatibility version mechanism provides an upgrade rollback path while at least one node is still on the previous version, or the compatibility version is still frozen at a previous version.
```

modules/ROOT/pages/cluster-compatibility-version.adoc
```diff
@@ -0,0 +1,221 @@
...
+
+
+== How the Cluster Compatibility Version Works
+
+Without a freeze, the cluster compatibility version is always the minimum version across all live nodes in the cluster.
+As you upgrade nodes one by one, the minimum advances when older nodes are replaced.
+
+{sgw} uses the compatibility version to gate cluster-wide feature activation.
+Features that could cause incompatibilities with older nodes remain inactive until the compatibility version reaches the version that introduced them.
+Node-local features, such as revision caching and import processing, activate immediately on each upgraded node regardless of the compatibility version.
```
bbrks: I am getting hung up on 'import processing' here mostly because import generally is **_not_** a node-local feature, but isn't gated because it has existed for such a long time

I also feel like we need to define `node-local` somewhere - since it is something that came out of Couchbase Server's implementation/architecture design. I would perhaps suggest something more descriptive, like:

```suggestion
New features that do not depend on metadata shared by other Sync Gateway nodes activate immediately on each upgraded node regardless of the compatibility version. These are considered 'node-local' features, and include areas such as caching, read-only operations, general bug fixes and performance improvements.
```

modules/ROOT/pages/cluster-compatibility-version.adoc
```diff
@@ -0,0 +1,221 @@
...
+[source,json]
+----
+{
+  "cluster_compat_version": "4.0",
+  "nodes": {
+    "node1-uid": "4.0",
+    "node2-uid": "4.0",
+    "node3-uid": "4.0"
+  },
+  "frozen_cluster_compat_version": "4.0"
```
bbrks: We should do these examples with 4.1 and 4.2 - it's not possible to view or freeze cluster compat version through 4.0 nodes since the feature only lands in 4.1

modules/ROOT/pages/cluster-compatibility-version.adoc
```diff
@@ -0,0 +1,221 @@
...
+Call xref:ROOT:rest_api_admin_static.adoc#tag/Server/operation/get__cluster_compat_version[GET /_cluster_compat_version] at any time to confirm the freeze is still active and to check the individual version of each node.
+
+.Step 4: Verify all nodes are on the new version
+
+Call xref:ROOT:rest_api_admin_static.adoc#tag/Server/operation/get__cluster_compat_version[GET /_cluster_compat_version] and confirm all nodes in the `nodes` map show the new version.
+
+[source,json]
+----
+{
+  "cluster_compat_version": "4.0",
```
bbrks: ditto - 4.1 and 4.2

bbrks (changes requested):
LGTM other than some minor suggestions

modules/ROOT/pages/cluster-compatibility-version.adoc
idulo: That's an important caveat, actually. Should be done as a Warning/Info banner or equivalent for users @iamfortune

modules/ROOT/pages/cluster-compatibility-version.adoc
```diff
@@ -0,0 +1,221 @@
...
+NOTE: Downgrading after the full cluster has been upgraded is not supported.
+The cluster compatibility version mechanism provides an upgrade rollback path while at least one node is still on the previous version, or the compatibility version is still frozen at a previous version.
+
+
+== How the Cluster Compatibility Version Works
+
+Without a freeze, the cluster compatibility version is always the minimum version across all live nodes in the cluster.
+As you upgrade nodes one by one, the minimum advances when older nodes are replaced.
+
+{sgw} uses the compatibility version to gate cluster-wide feature activation.
```
idulo, line 35: Should we be making notes going forward as to which features are cluster compatible depending on SGW version? Or saying that no features are specifically marked as cluster compatible in 4.1 at the moment @bbrks

idulo (changes requested):

modules/ROOT/pages/cluster-compatibility-version.adoc
bbrks: Yeah that's a good point - In 4.1 the only cluster compat version gated feature is metadata migration.

We'll probably have to do that going forwards so customers know which features they can't test behind a frozen upgrade.
There are definitely going to be cases where we have CCV-gated changes that aren't customer-facing features (e.g. internal metadata format changes, etc.) so maybe alongside the feature documentation itself makes sense as a place for this info?

bbrks (dismissed):

bbrks (approved):

PR #995, opened 2026-07-15
DoC-13337-add-special-characters-data-modeling

iamfortune:
Docs Issue: [DOC-13337](https://jira.issues.couchbase.com/browse/DOC-13337?atlOrigin=eyJpIjoiOTQzNDE3ODUyYzhlNDBiNmE2ZGEwOTA2YWJkMzNiMmYiLCJwIjoiaiJ9)

PR to add special characters to SGW data modelling page 

Preview URL:
https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-13337-add-special-characters-data-modelling

[DOC-13337]: https://couchbasecloud.atlassian.net/browse/DOC-13337?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/ROOT/pages/data-modeling.adoc
+++ b/modules/ROOT/pages/data-modeling.adoc
@@ -96,6 +96,54 @@ As a result, documents can represent complex data structures in a readily parsab
 These attachments provide a means to store large media files or any other non-textual data.
 Couchbase Lite supports attachments of unlimited size, although the Sync Gateway imposes a 20MB limit for attachments synced to it.
 
+[#special-characters]
+=== Special Characters
+
+Sync Gateway automatically escapes certain characters in any string property
+name or property value within a document body, including values nested
+inside objects or arrays. This behavior is uncommon and only affects
+documents that contain these specific characters.
+
+Sync Gateway escapes the characters as follows:
+
+* `<` becomes `\u003c`
+* `>` becomes `\u003e`
+* `&` becomes `\u0026`
+
+.Special character escaping example
+====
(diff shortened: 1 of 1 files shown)


iamfortune:
@bbrks I have a few questions related to this PR

1. Does this escaping apply to the document `_id`, or only to property names/values inside the JSON body? This determines whether the note stays scoped to the "Value" subsection or needs to touch the "Key" section too.

2. Does escaping apply to property names (JSON keys) as well as values, or values only? The wording "property name or value" in my draft is only correct if this is confirmed either way.

3. Is `<`, `>`, `&` the complete list, or are other characters (`"`, `\`, non-ASCII) also escaped? If the list is incomplete, the docs teach readers to expect a subset of what actually happens.

4. When a client reads the document back, does Sync Gateway return it still escaped, or does it unescape automatically before returning? This decides whether the "use a JSON deserializer or match the escaped form" guidance is accurate or backwards.

5. Does this escaping apply identically across all three write paths (Couchbase Lite SDKs, Sync Gateway REST APIs, Couchbase Server SDKs with shared bucket access), or are there exceptions per path? I reused this list from Property Naming without confirmation it holds here too.

bbrks:
I did try to answer these questions quickly a couple of weeks ago when I saw this but didn't have any of the context of the original issue loaded back into my head to confidently answer them.

Will take another pass on this whilst on support duty.

PR #1000, opened 2026-07-17
DOC-14114-stes-sgw-containers

iamfortune:
Docs Issue: [](https://jira.issues.couchbase.com/browse/DOC-14114)

PR to update how to run containers for SGW with docker 

Preview URL:
https://preview.docs-test.couchbase.com/docs-sync-gateway-DOC-14114-stes-sgw-containers

### DIFF
--- a/modules/start-here/pages/get-started-install-docker.adoc
+++ b/modules/start-here/pages/get-started-install-docker.adoc
@@ -114,12 +114,19 @@ For production deployments, enable TLS and use `couchbases://cb-server` instead.
 
 ====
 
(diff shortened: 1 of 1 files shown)


bbrks:
I left a comment on the Jira ticket since I don't think we want to actually go this route to solve the issue... it was already being tracked to fix the default container config

bbrks (changes requested):
See above - maybe we do document this for _current versions_ of SG affected by the bad default config, but still need to get the fix in and not describe this as authoritatively.

iamfortune:
Thank you for this, I'll close this ticket, also closed the ticket on jira.

